PR #1, opened 2023-12-07
Add Security Considerations Section

cdekater:


jiceatscion (approved):
...assuming typo is fixed.

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,101 @@ This section describes the steps that a SCION egress border router must perform
...
 
-TODO security considerations.
+This section describes the possible security risks and attacks that SCION's data plane may be prone to, and how these attacks may be mitigated. It first discusses security risks that pertain to path authorization, followed by a section on all other data-plane security risks.
+
+**Note:** This section only discusses SCION data plane- and forwarding-specific security considerations. For security considerations related to the SCION control-plane PKI, see {{I-D.scion-cppki}}. {{I-D.scion-cp}} includes security considerations that concern the SCION control plane and routing.
+
+
+## Security Risks in Regard to Path Authorization
+
+A central property of the SCION path-aware data plane is path authorization. Path authorization guarantees that data packets always traverse the network along paths segments authorized in the control plane by all on-path ASes. This section discusses how an adversary may attempt to violate the path-authorization property, as well as SCION's answers to these attacks. The following attacks are imaginable:
```
knmeynell, line 1365:
```suggestion
A central property of the SCION path-aware data plane is path authorization. Path authorization guarantees that data packets always traverse the network along path segments authorized in the control plane by all on-path ASes. This section discusses how an adversary may attempt to violate the path-authorization property, as well as SCION's answers to these attacks. The following attacks are imaginable:
```

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,101 @@ This section describes the steps that a SCION egress border router must perform
...
+
+- Crafting unauthorized hop fields (see [](#crafting)),
+- modifying the validity period of a path segment (see [](#mod-validity)), or
+- path splicing, that is, constructing an unauthorized segment by combining hop fields of different valid segments (see [](#path-splicing)).
+
+
+### Crafting Unauthorized Hop Fields {#crafting}
+
+Hop fields are protected with MACs. To be able to craft and misuse unauthorized hop fields, an adversary must therefore either determine the MAC's key or try to break the MAC directly.
```
knmeynell, line 1374: Is 'crack' a better work than 'break' here (and further down the text)?

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,101 @@ This section describes the steps that a SCION egress border router must perform
...
+
+Modifications of the SCION address header can be discovered by the destination endpoint if the packet is integrity protected (which is currently not supported by SCION).
+
+
+### Off-Path Attacks - Denial of Service
+
+An off-path adversary is located outside of the path between source endpoint and destination endpoint. In SCION, the source endpoint selects the forwarding path in the control plane, and includes it in the header of the data packet to be forwarded. This limits the abilities of an off-path adversary to influence the forwarding in the date plane. The adversary can merely attempt to disrupt the connectivity of the chosen path and force the source endpoint to select a new path. One way to do this is by flooding a link on the path with excessive traffic; that is, performing a volumetric denial of service attack. However, this does not stop the data forwarding: As SCION supports multipath, it is still possible to switch to another, non-congested path. Only in the case of a disruptive last-mile link for which no alternative exists, the data forwarding will be severely frustrated.
```
knmeynell, line 1433:
```suggestion
An off-path adversary is located outside of the path between source endpoint and destination endpoint. In SCION, the source endpoint selects the forwarding path in the control plane, and includes it in the header of the data packet to be forwarded. This limits the abilities of an off-path adversary to influence the forwarding in the data plane. The adversary can merely attempt to disrupt the connectivity of the chosen path and force the source endpoint to select a new path. One way to do this is by flooding a link on the path with excessive traffic; that is, performing a volumetric denial of service attack. However, this does not stop the data forwarding: As SCION supports multipath, it is still possible to switch to another, non-congested path. Only in the case of a disruptive last-mile link for which no alternative exists, the data forwarding will be severely frustrated.
```

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,101 @@ This section describes the steps that a SCION egress border router must perform
...
+
+SCION also provides protection against so-called reflection-based attacks. Here, the adversary sends high amounts of request packets to a server, but forges the source address of the packets, by replacing its own address with the address of the victim. As servers are unable to distinguish legitimate from spoofed requests, they reply directly to the victim, thus exhausting the connection to the victim. In SCION, however, packets contain the path in their packer header, and response packets are simply returned along this contained path to the actual sender. SCION thus prevents from this kind of DoS attacks.
```
knmeynell, line 1435:
```suggestion
SCION also provides protection against so-called reflection-based attacks. Here, the adversary sends high amounts of request packets to a server, but forges the source address of the packets by replacing its own address with the address of the victim. As servers are unable to distinguish legitimate from spoofed requests, they reply directly to the victim, thus exhausting the connection to the victim. In SCION, however, packets contain the path in their packer header, and response packets are simply returned along this contained path to the actual sender. SCION thus prevents from this kind of DoS attacks.
```

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,101 @@ This section describes the steps that a SCION egress border router must perform
...
+
+**Note** that SCION does not protect against two other tyoes of DoS attacks, namely transport protocol attacks and application layer attacks. In the case of a transport protocol attack, the adversary opens a large number of connections in order to exhaust resources on the target server. If an adversary performs an application layer attack, it overloads or crashes the target application with a large volume of or specially crafted application-layer requests, such as HTTP floods. As these attacks take place in the transport layer or application layer, respectively, and not the network layer, they are out of SCION's scope.
```
knmeynell, line 1437:
```suggestion
**Note** that SCION does not protect against two other types of DoS attacks, namely transport protocol attacks and application layer attacks. In the case of a transport protocol attack, the adversary opens a large number of connections in order to exhaust resources on the target server. If an adversary performs an application layer attack, it overloads or crashes the target application with a large volume of, or specially crafted application-layer requests such as HTTP floods. As these attacks take place in the transport layer or application layer rather than the network layer, they are out of SCION's scope.
```

draft-dekater-scion-dataplane.md
nicorusti: I'd say "crack the underlying MAC algorithm"

nicorusti, line 1374:
```suggestion
Hop fields are protected with MACs. To be able to craft and misuse unauthorized hop fields, an adversary must therefore either determine the MAC's key or crack the underlying MAC algorithm.
```

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,101 @@ This section describes the steps that a SCION egress border router must perform
...
+- Crafting unauthorized hop fields (see [](#crafting)),
+- modifying the validity period of a path segment (see [](#mod-validity)), or
+- path splicing, that is, constructing an unauthorized segment by combining hop fields of different valid segments (see [](#path-splicing)).
+
+
+### Crafting Unauthorized Hop Fields {#crafting}
+
+Hop fields are protected with MACs. To be able to craft and misuse unauthorized hop fields, an adversary must therefore either determine the MAC's key or crack the underlying MAC algorithm.
+
+To determine an unknown MAC key, the adversary may perform a brute-force attack. Candidate keys can be validated by checking the MAC contained in sample hop fields. But as SCION uses 128-bit keys by default, such an off-line attack is computationally infeasible in practice. Another obstacle to adversaries is the short life of the keys for MAC computation, with a validity period of 24 hours.
```
nicorusti, line 1376: IN this draft we don't say much about the AS forwarding key. We don't say that the key should be 128 bit, and we also don't explain what the assumptions and threat model is (e.g. that  the key is shared across all routers within an AS). We should clarify this.
Also, we don't describe how the key is rotated

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,101 @@ This section describes the steps that a SCION egress border router must perform
...
+
+The destination endpoint may only discover such path header modifications if some kind of packet integrity protection is in place. In this case, the receiving endpoint verifies that a packet, including its path, is the same as the one sent by the source endpoint. However, SCION currently does not provide this type of protection.
+
+Moreover, packet integrity protection is not enough if there are two colluding adversaries on the path. These colluding adversaries can forward the packet between them using a different path than selected by the source endpoint: The first on-path attacker remodels the packet header arbitrarily, and the second on-path attacker changes the path back to the original source-selected path, such that the integrity check by the destination endpoint succeeds. To prevent this attack and to defend against multiple on-path adversaries in general, path validation is required (which is currently not supported by SCION).
```
nicorusti, line 1421: Is this just the wormhole attack, as described in the CP draft?

nicorusti, line 1421:
```suggestion
Moreover, packet integrity protection is not enough if there are two colluding adversaries on the path. These colluding adversaries can forward the packet between them using a different path than selected by the source endpoint: The first on-path attacker remodels the packet header arbitrarily, and the second on-path attacker changes the path back to the original source-selected path, such that the integrity check by the destination endpoint succeeds. To prevent this attack and to defend against multiple on-path adversaries in general, proof of transit is required (which is offered by an experimental extension).
```

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,101 @@ This section describes the steps that a SCION egress border router must perform
...
+
+In all cases, *if* an online brute-force attack succeeds, the obtained hop field can be reused until its eventual expiration.
+
+
+### Modifying the Validity Period {#mod-validity}
+
+The metadata for each path segment in a SCION path is stored in the corresponding SCION packet header. It includes a timestamp, which is set by the initiator of the PCB. The expiry time of the hop fields within a path segment is calculated relative to this timestamp (see also [](#scion-path-type)).
+
+An adversary could try to manipulate the timestamp, in order to modify the validity period of the hop fields within a path segment. However, SCION offers a protection against this kind of attack: As both the timestamp and the expiry time build the input for the calculation of a hop field's MAC, an unauthorized change of the timestamp will invalidate the MAC, and will therefore not pass unnoticed.
```
nicorusti, line 1389:
```suggestion
An adversary could try to manipulate the timestamp, in order to modify the validity period of the hop fields within a path segment. However, SCION offers a protection against this kind of attack: As both the timestamp and the expiry time build the input for the calculation of a hop field's MAC, an unauthorized change of the timestamp will invalidate the MAC, and will therefore be detected and render the hop field unusable for forwarding.
```
Correct @matzf ?

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,97 @@ This section describes the steps that a SCION egress border router must perform
...
+
+
+#### Modification of the Address Header
+
+Besides modifying the SCION path header, an on-path adversary can also modify the source and destination addresses in the SCION address header. If an adversary changes the address of the destination endpoint, it needs to modify the "future" path between itself and the new destination accordingly; otherwise the packet would simply be dropped by the last AS on the path.
+
+Modifications of the SCION address header can be discovered by the destination endpoint if the packet is integrity protected (which is currently not supported by SCION).
```
nicorusti, line 1424:
```suggestion
Modifications of the SCION address header can be discovered by the destination endpoint if the packet is integrity protected with the SPAO extension. Currently, SCION as described in this document does not provide this type of protection.
```

SPAO also MACs the endpoint address, right? (Yes, according to [docs](https://docs.scion.org/en/latest/protocols/authenticator-option.html#authenticated-data)).

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,97 @@ This section describes the steps that a SCION egress border router must perform
...
+
+
+### Off-Path Attacks - Volumetric Denial of Service
+
+SCION's path-awareness limits the abilities of an off-path adversary to influence forwarding in the data plane. An adversary can merely attempt to disrupt the connectivity of the chosen path choosen by an endpoint, and force the source endpoint to select a new path. One way to do this is by flooding a link on the path with excessive traffic; that is, performing a volumetric denial of service attack. Should this happen, the endpoint cna switch to another, non-congested path thanks to multipath, as long as any alternative inter-domain path is available.
+
+SCION also provides protection against so-called reflection-based attacks. Here, the adversary sends high amounts of request packets to a server, but forges the source address of the packets with the address of the victim, triggering the server to send a reply to the victim. In SCION, however, packets contain the path in their packer header, and response packets are simply returned along this contained path to the actual sender. SCION thus prevents from this kind of DoS attacks.
```
nicorusti, line 1431: Is this also valid for vanilla SCION? Could the same attack happen if the attacker is in the same AS as the victim? 
(So, in this case, the attack surface would be greatly reduced, nut it would only be eliminated with SPAO)?

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,97 @@ This section describes the steps that a SCION egress border router must perform
...
+
+**Note** that SCION does not protect against two other types of DoS attacks, namely transport protocol attacks and application layer attacks. Such attacks are out of SCION's scope. However, the additional information contained in the SCION header enables more targeted filtering, e.g., by ISD, AS or path length.
```
nicorusti, line 1433: Does the wording sound reasonable?

draft-dekater-scion-dataplane.md
matzf: True. I think the SPAO was left out intentionally. Makes sense to allude to its existence, but maybe it would be good to be a bit more generic.

> Modifications of the SCION address header can be discovered by the destination endpoint by a data integrity protection system. Such a data integrity protection system, losely analogous to the IPSec Authentication Header, exists for SCION but is out of scope for this document.

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,83 @@ This section describes the steps that a SCION egress border router must perform
...
+
+
+### Modifying the Validity Period {#mod-validity}
+
+The metadata for each path segment in a SCION path is stored in the corresponding SCION packet header. It includes a timestamp, which is set by the initiator of the PCB. The expiry time of the hop fields within a path segment is calculated relative to this timestamp (see also [](#scion-path-type)).
+
+An adversary could try to manipulate the timestamp, in order to modify the validity period of the hop fields within a path segment. However, as both the timestamp and the expiry time build the input for the calculation of a hop field's MAC, an unauthorized change of the timestamp will invalidate the MAC, and will therefore be detected and render the hop field unusable for forwarding..
```
matzf, line 1387:
```suggestion
A central property of the SCION path-aware data plane is path authorization. Path authorization guarantees that data packets always traverse the network along path segments authorized in the control plane by all on-path ASes. This section discusses how an adversary may attempt to violate the path-authorization property, as well as SCION's prevention mechanisms to these attacks; either an attacker can attempt to create unauthorized hop fields, or they can attempt to create illegitimate paths assembled from authentic individual hop fields.

The main protection mechanism here is the hop field MAC ( see ... ), authenticating the hop field content, consisting of ingress/egress interface identifiers, creation and expiration timestamp and, virtually, the preceding hop field MACs in the path segment. 
Recall that the MAC is computed using the issuing AS's secret "forwarding key" which is shared across the SCION routers and control plane services (only!) within this AS.

### Forwarding key compromise

For the current default MAC algorithm, AES-CMAC truncated to 48 bits, key recovery attacks from (any number of) known plaintext/MAC combinations is computationally infeasible, as far as is publicly known. 
In addition, MAC algorithm can be freely chosen by each ASes, enabling algorithmic agility for MAC computations. Should a MAC algorithm be discovered to be weak or insecure, each AS can quickly switch to a secure algorithm without the need for coordination with other ASes.

A more realistic risk to the secrecy of the forwarding key is exfiltration from a compromised router or control plane service.
An AS can optionally rotate its forwarding key at regular intervals to limit the exposure after a temporary device compromise. However, as is perhaps self-evident, such a key rotation scheme cannot mitigate the impact of an undiscovered, permanent compromise of a device.

### Forging hop field MAC

As a second method to break the path authorization is to directly forge a hop field in an online attack, using the router as an oracle to determine the validity of the hop field MAC. The adversary needs to send one packet per guess for verification. For a 6-byte MAC, the adversary would need an expected 2<sup>47</sup> (~140 trillion) tries to successfully forge the MAC of a single hop field.
As the router only checks MACs during the encoded validity period of the hop field, which is limited by the packet header format to 24 hours, these tries need to occur in a limited time period. This results in a seemingly infeasible number of ~1.6e9 guesses per second.
In the unlikely case that an online brute-force attack succeeds, the obtained hop field can be used until its inevitable expiration after the just mentioned 24 hour limit.

```

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,83 @@ This section describes the steps that a SCION egress border router must perform
...
+
+SCION also provides protection against so-called reflection-based attacks. Here, the adversary sends high amounts of request packets to a server, but forges the source address of the packets with the address of the victim, triggering the server to send a reply to the victim. In SCION, however, packets contain the path in their packet header, and response packets are simply returned along this contained path to the actual sender. SCION thus prevents from this kind of DoS attacks.
+
+**Note** that SCION does not protect against two other types of DoS attacks, namely transport protocol attacks and application layer attacks. Such attacks are out of SCION's scope. However, the additional information contained in the SCION header enables more targeted filtering, e.g., by ISD, AS or path length.
```
matzf, line 1419:
```suggestion
## Off-Path Attacks

SCION's path-awareness limits the abilities of an off-path adversary to influence forwarding in the data plane. Once a packet is "in flight", it will follow its set route, no matter what an adversary may do.
An adversary can attempt to disrupt the connectivity of said path by flooding a link with excessive traffic (see [](#dos) below). After detecting the congestion, the endpoint can switch to another, non-congested path for subsequent packets.
Instead of actually disrupting the network connectivity, an off-path adversary can also attempt to influence the path selection of the endpoint by fraudulently signalling a network interruption with forged SCMP messages. This attack scenario, and protection mechanisms, will be discussed in a separate document on the SCMP.

## Volumetric Denial of Service Attacks {#dos}

An adversary can attempt to disrupt the connectivity of a network path by flooding a link with excessive traffic. In this case, the endpoint can switch to another, non-congested path for subsequent packets.

SCION provides protection against certain reflection-based DoS attacks. Here, the adversary sends requests to a server with the source address set to the address of the victim. The server will send a reply, typically larger than the request, to the victim. As long as the attacker and the victim are located in different ASes, this can be prevented in SCION. The reply packets are simply returned along reversed path to the actual sender, regardless of the source address information. Thus, the reflected will be forwarded to the attackers AS (where it will be discarded because the destination AS does not match).

On the flip side, the path choice of the endpoint may possibly be exploited by an attacker to create intermittent congestion with relatively low send rate; the attacker can abuse the latency differences of the available paths, sending at precisely timed intervals to cause short, synchronized bursts of packets near the victim.

**Note** that SCION does not protect against two other types of DoS attacks, namely transport protocol attacks and application layer attacks. Such attacks are out of SCION's scope. However, the additional information contained in the SCION header enables more targeted filtering, e.g., by ISD, AS or path length.
```

draft-dekater-scion-dataplane.md
```diff
@@ -1342,20 +1355,87 @@ This section describes the steps that a SCION egress border router must perform
...
+
+
+## Off-Path Attacks
+
+SCION's path-awareness limits the abilities of an off-path adversary to influence forwarding in the data plane. Once a packet is "in flight", it will follow its set route, no matter what an adversary may do.
+An adversary can attempt to disrupt the connectivity of said path by flooding a link with excessive traffic (see [](#dos) below). After detecting the congestion, the endpoint can switch to another, non-congested path for subsequent packets.
+Instead of actually disrupting the network connectivity, an off-path adversary can also attempt to influence the path selection of the endpoint by fraudulently signalling a network interruption with forged SCMP messages. This attack scenario, and protection mechanisms, will be discussed in a separate document on the SCMP.
```
nicorusti, line 1413: @matzf We don't talk about this kind of SCMP message, I'd therefore omit this line and keep it for the security considerations of SCMP, once we get there

PR #15, opened 2024-05-21
Clarify implications of AS forwarding key compromise

### LINKED ISSUE #12: Clarify implications of AS forwarding key compromise
> From:	Joel Halpern
> Before that, I would also ask them for the data plane document for an analysis of what the implications are if a key shared by all the SCION routers of an AS is compromised.  It is ahrd to tell if the design assumptions are workable without that. 

 We should better clarify this, we don't say what an attacker can do if an attacker gets the AS forwarding key. 
 Current text is: 
 ```
5.1.1.  Forwarding key compromise

   For the current default MAC algorithm, AES-CMAC truncated to 48 bits,
   key recovery attacks from (any number of) known plaintext/MAC
   combinations is computationally infeasible, as far as publicly known.
   In addition, the MAC algorithm can be freely chosen by each AS,
   enabling algorithmic agility for MAC computations.  Should a MAC
   algorithm be discovered to be weak or insecure, each AS can quickly
   switch to a secure algorithm without the need for coordination with
   other ASes.

   A more realistic risk to the secrecy of the forwarding key is
   exfiltration from a compromised router or control plane service.  An
   AS can optionally rotate its forwarding key at regular intervals to
   limit the exposure after a temporary device compromise.  However, as
   is perhaps self-evident, such a key rotation scheme cannot mitigate
   the impact of an undiscovered, permanent compromise of a device.
   ```

matzf:
Fixes #12


draft-dekater-scion-dataplane.md
```diff
@@ -1374,6 +1374,11 @@ In addition, the MAC algorithm can be freely chosen by each AS, enabling algorit
 A more realistic risk to the secrecy of the forwarding key is exfiltration from a compromised router or control plane service.
 An AS can optionally rotate its forwarding key at regular intervals to limit the exposure after a temporary device compromise. However, as is perhaps self-evident, such a key rotation scheme cannot mitigate the impact of an undiscovered, permanent compromise of a device.
 
+When an AS's forwarding key is compromised, an attacker can forge hop field MACs, undermining path authorization. As path segments are only checked for validity and policy compliance during path discovery during normal operation, and routers only perform minimal sanity checks of hop fields with valid MACs, creating fraudulent hop fields with valid MACs allows an attacker to bypass most path segment validity checks. Thus, an attacker can create path segments that violate the AS's local policy and/or general path segment validity requirements.
```
jiceatscion, line 1377: > As path segments are only checked for validity and policy compliance during path discovery during normal operation, and routers only perform minimal sanity checks of hop fields with valid MACs, creating fraudulent hop fields with valid MACs allows an attacker to bypass most path segment validity checks.

Speaking that one requires breathing exercises. Any way to split it? (yeah, me, the lumper :-) )

jiceatscion (changes requested):

draft-dekater-scion-dataplane.md
matzf: Done, but I felt I had to combine the next sentences to make it work.
Better?

jiceatscion: Yes.

jiceatscion (approved):

draft-dekater-scion-dataplane.md
```diff
@@ -1374,6 +1374,11 @@ In addition, the MAC algorithm can be freely chosen by each AS, enabling algorit
 A more realistic risk to the secrecy of the forwarding key is exfiltration from a compromised router or control plane service.
 An AS can optionally rotate its forwarding key at regular intervals to limit the exposure after a temporary device compromise. However, as is perhaps self-evident, such a key rotation scheme cannot mitigate the impact of an undiscovered, permanent compromise of a device.
 
+When an AS's forwarding key is compromised, an attacker can forge hop field MACs, undermining path authorization. Recall that path segments are only checked for validity and policy compliance during the path discovery phase, and routers only perform minimal sanity checks of hop fields with valid MACs. Consequently, creating fraudulent hop fields with valid MACs allows an attacker to bypass most path segment validity checks, and to create path segments that violate the AS's local policy and/or general path segment validity requirements.
```
nicorusti, line 1377: >  Recall that path segments are only checked for validity and policy compliance during the path discovery phase, and routers only perform minimal sanity checks of hop fields with valid MACs.

I find the language here a bit confusing, because hop field MAC validity is checked at forwarding by the DP, and this is more than _ minimal sanity checks_. I'd just mention that as the DP is only verifying MACs in hop fields, a forget MAC allows an attacker to..

matzf: The structure of the sentence I have adapted. Better?

draft-dekater-scion-dataplane.md
```diff
@@ -1374,6 +1374,11 @@ In addition, the MAC algorithm can be freely chosen by each AS, enabling algorit
...
+In particular, an attacker with access to the forwarding key of at least two compromised ASes could create paths that include loops (limited by the maximum number of hop fields of a path).
+
+While a forwarding key compromise and the resulting loss of path authorization is a serious degradation of SCION's routing security properties, this does not affect, for example, access control or data security for the hosts in the affected AS.
```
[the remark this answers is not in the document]
matzf: Added another sentence to clarify that this does not affect routing for legitimate senders.

True, the attacker will most likely try to use this to splice parts of path segments, but it turns out this is quite involved because of the MAC chaining. I tried to explain this as short as possible.

nicorusti (approved):
Sorry for the delay, thank you for your changes, looks good!

PR #16, opened 2024-05-23
Remove forward references to SCMP

### LINKED ISSUE #8: SCMP - Be careful about forward references
>  SCION also includes its own protocol to communicate failures to endpoints,
   the SCION Control Message Protocol (SCMP).  This protocol will be
   described in a separate document, which will follow later.

Text like this doesn't stand on its own.

jiceatscion:
Fixes #8

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -787,8 +787,8 @@ The SCION Hop-by-Hop Options and End-to-End Options headers have the following f
 
 
 - `NextHdr`: Unsigned 8-bit integer. Identifies the type of header immediately following the Hop-by-Hop/End-to-End Options header. Values of this field respect the Assigned SCION Protocol Numbers (see also [](#protnum)).
-- `ExtLen`: Unsigned 8-bit integer. The value of this field is computed as the length of the complete Hop-by-Hop/End-to-End Options header in multiples of 4-bytes minus 1.
-- `Options`: This is a variable-length field. The length of this field MUST be such that the complete length of the Hop-by-Hop/End-to-End Options header is an integer multiple of 4 bytes. The `Options` field contains one or more Type-Length-Value (TLV) encoded options. For details, see [](#optfld).
+- `ExtLen`: Unsigned 8-bit integer. The value of this field is computed as the length of the header in 4-octet units not including the first 4 octets. That is: `ExtLen = uint8(((options_len + 2) / 4) - 1)` where `options_len + 2` is a multiple of 4.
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-dataplane.md
```diff
@@ -204,7 +204,7 @@ SCION leverages source-based path selection, where path information is embedded
 
 **Path Transparency**: Path transparency is a property of a network architecture that gives endpoints full visibility over the network paths their packets are taking. Path transparency is weaker than path control.
 
-**SCMP**: SCION Control Message Protocol. SCMP is used for signaling connectivity problems, analogous to the Internet Control Message Protocol (ICMP). SCMP provides network diagnostic and error messages.
+**SCMP**: Future SCION Control Message Protocol. SCMP is intended to be analogous to the Internet Control Message Protocol (ICMP).
```
nicorusti, line 207:
```suggestion
```

draft-dekater-scion-dataplane.md
```diff
@@ -1445,24 +1442,25 @@ SCION attempts to take the IANA's assigned Internet protocol numbers into consid
 
 The protocol numbers are used in the SCION header to identify the next level protocol.
 
+SCION reserves a number for a future protocol: the SCION Control Message Protocol (SCMP). Support for this protocol is optional and it has not yet been formally specified. A work-in-progress specification exists: [the scion.org SCMP specification document](https://docs.scion.org/en/latest/protocols/scmp.html).
```
nicorusti, line 1445: Maybe we could remove this comment and leave it in the table? I think we have the same problem for DrKey somewhere else

nicorusti (changes requested):

draft-dekater-scion-dataplane.md
jiceatscion: I can't figure what you're suggesting. The review UI seems to say that you want to remove the new text entirely as well as the old text. Is that what you're suggesting?

draft-dekater-scion-dataplane.md
jiceatscion: I was acting on Eliot's suggestion: mention the protocol informally. If we just leave in the table with no explanation, we *will* get the question "what is this?". What is the motivation for your suggestion?

nicorusti (approved):
Looks good, I'll ask Eliot if the reference works, If I don't hear back in a few days I'd berge and just leave the corresponding issue open

nicorusti:
Eliot gave us a thumbs up for the current reference, merged.

PR #18, opened 2024-05-23
Add service field table

### LINKED ISSUE #13: Clarify service field
We currently document type service 0b0100, but we don't further document possible values for it. We might need to add another table with potential values for the service field, as defined in [here](https://github.com/scionproto/scion/blob/master/pkg/addr/svc.go)
```
                      Table 2: Address length values

    +================+======+=======================+================+
    | Length (bytes) | Type |  Type/Length (binary) | Interpretation |
    +================+======+=======================+================+
    | 4              | 0    | 0b0000                | IPv4           |
    +----------------+------+-----------------------+----------------+
    | 4              | 1    | 0b0100                | Service        |
    +----------------+------+-----------------------+----------------+
    | 16             | 0    | 0b1100                | IPv6           |
    +----------------+------+-----------------------+----------------+
    | other          |      |                       | Unassigned     |
    +----------------+------+-----------------------+----------------+
   
  ```

Relates to https://github.com/scionassociation/scion-cp_I-D/issues/15

### PR DESCRIPTION
**Pull request by jiceatscion**
Fixes #13

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -163,6 +163,8 @@ The SCION data plane fundamentally differs from today's IP-based data plane in t
 
 SCION leverages source-based path selection, where path information is embedded in the packet header - this is called packet-carried forwarding state (PCFS). This section explains how data packets are forwarded through the network, how the SCION inter-domain routing differs from intra-domain routing, and how endpoints can construct end-to-end paths from path segments. It also briefly touches the concept of path authorization, which ensures that data packets always traverse the network along authorized paths.
 
+The data transmission order for SCION is the same as for IPv6 as defined in Introduction of {{RFC8200}}.
+
 **Note:** This is the very first version of the SCION Data Plane document. We are aware that the draft is far from perfect, and hope to improve the content in later versions of the document. To reach this goal, any feedback is welcome and much appreciated. Thanks!
 
 **Note:** It is assumed that readers of this draft are familiar with the basic concepts of the SCION next-generation inter-domain network architecture. If not, please find more detailed information in the IETF Internet Drafts {{I-D.scion-overview}}, {{I-D.scion-components}}, {{I-D.scion-cppki}}, and {{I-D.scion-cp}}, as well as in {{CHUAT22}}, especially Chapter 2. A short description of the SCION basic terms and elements can be found in [](#terms) below.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by nicorusti — COMMENTED** [review]
Thanks for adding the table. I think we need to have a more comprehensive discussion on service discovery and wether we add this to the draft or not. 
This because in some way,  just adding this extra table will raise more questions..

**Comment by jiceatscion** [comment]
This specifically addresses: https://github.com/scionassociation/scion-dp_I-D/issues/13. Should we then close it as "won't fix"?

**Comment by jiceatscion** [comment]
I added a brief description of what a service address is. That should clarify the reference to it that is in the configuration PR.

File: draft-dekater-scion-dataplane.md [hunk 1]
```diff
@@ -447,8 +449,18 @@ The SCION common header has the following packet format:
 | other          |      |                      | Unassigned     |
 {: #table-3 title="Allocations of type values to length values"}
 
-- `RSV`: These bits are currently reserved for future use.
+A service address designates a set of endpoint addresses rather than a singular one. A packet addressed to a service is redirected to any one endpoint-addresses that is known to be part of the set.
```
nicorusti, line 452: Based on table 3, It seems that an address is a service address if either `DT/DL`=1/0 or if `ST/SL`=1/0. Is source ever used? If not, it should be clearer. Maybe we could just say, before the service address table, that those addresses are for case  `DT/DL`=1/0

draft-dekater-scion-dataplane.md
```diff
@@ -447,8 +449,18 @@ The SCION common header has the following packet format:
...
+Current known values are:
 
+| Hexadecimal value | Short Name | Description            |
+|-------------------+------------+------------------------|
+| 0x00000001        | DS         | Discovery Service      |
+| 0x00000002        | CS         | Control Service        |
+| 0x0000ffff        | None       | Reserved invalid value |
+{: #table-4 title="Known Service values"}
+
+- `RSV`: These bits are currently reserved for future use.
```
nicorusti, line 463: Would it make more sense to place the overall service addresses table one section below, at the end of parargaph `## Address Header` ? 

I am saying this because service addresses are special values of `DstHostAddr` when `DT/DL`=1/0, this field is in the address header, not in the common header

nicorusti (changes requested):

draft-dekater-scion-dataplane.md
jiceatscion: Made it more obvious that this goes for both src and dst.

draft-dekater-scion-dataplane.md
jiceatscion: Done.

nicorusti (approved):
LGTM

PR #21, opened 2024-05-28
Add information about configuration. Esp. Interface configuration.

jiceatscion:
Addresses https://github.com/scionassociation/scion-cp_I-D/issues/17

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -186,7 +186,7 @@ SCION leverages source-based path selection, where path information is embedded 
 
 **Info Field (INF)**: Each path-segment construction beacon (PCB) contains a single info field, which provides basic information about the PCB. Together with hop fields (HFs), info fields are used to create forwarding paths.
 
-**Interface Identifier (Interface ID)**: 16 bit identifier that designates a SCION interface at the end of a link connecting two SCION ASes. Each interface belongs to one border router. Hop fields describe the traversal of an AS by a pair of interface IDs (the ingress and egress interfaces). The Interface ID MUST be unique within each AS, with the exception of ID 0 that denotes a connection between a border router and the AS's internal network. Each border router manages an instance of interface 0.
+**Interface Identifier (Interface ID)**: 16 bit identifier that designates a SCION interface at the end of a link connecting two SCION ASes. Each interface belongs to one border router. Hop fields describe the traversal of an AS by a pair of interface IDs (the ingress and egress interfaces). The Interface ID MUST be unique within each AS. Interface ID 0 is not a valid identifier, implementations can use it as the "unspecified" value.
 
 **Isolation Domain (ISD)**: In SCION, autonomous systems (ASes) are organized into logical groups called isolation domains or ISDs. Each ISD consists of ASes that span an area with a uniform trust environment (e.g., a common jurisdiction). A possible model is for ISDs to be formed along national boundaries or federations of nations.
 
@@ -254,19 +254,15 @@ The full forwarding process for a packet transiting an intermediate AS consists 
 
 ### Configuration
 
-Border routers require mappings from SCION  interface IDs to underlay addresses. Such information must be supplied to each router in an out of band fashion (e.g in a configuration file). For each link to a neighbor, these values must be configured:
+Border routers require mappings from SCION  interface IDs to underlay addresses. Such information must be supplied to each router in an out of band fashion (e.g in a configuration file). For each link to a neighbor, these values must be configured. A typical implementation will require:
 
-- Local interface ID
-- Neighbor type (core, parent, child, peer), depending on link type. Link type depends on mutual agreements between the organizations operating the ASes at each end of each link.
-- Neighbor ISD-AS number
-- Neighbor interface underlay address
+- Interface ID.
+- Link type (core, parent, child, peer). Link type depends on mutual agreements between the organizations operating the ASes at each end of each link.
+- Neighbor ISD-AS number.
+- For the router that manages the interface: the neighbor interface underlay address.
+- For the routers that do not manage the interface:  "intra-protocol" address of the router that does.
 
-In addition, each router must be configured with a list of interface IDs in use within the AS, containing:
-
-- Interface ID
-- "intra-protocol" address of the remote router
-
-In order to forward traffic to service endpoint addresses (`DT/DS` == 0b01) in the [common header](#common-header), a border router must translate service numbers into concrete host addresses. The method used to accomplish the translation is not defined by this document. It only depends on the implementation and the choices of each AS's administrator. In current practice this is accomplished by way of a configuration file. There are currently only two services addressed in this manner. As a result it is practical to maintain a static mapping in a configuration file. Other methods have been used, including DNS.
+In order to forward traffic to service endpoint addresses (`DT/DS` == 0b01 in the [common header](#common-header)), a border router must translate service numbers into concrete host addresses. The method used to accomplish the translation is not defined by this document. It only depends on the implementation and the choices of each AS's administrator. In current practice this is accomplished by way of a configuration file. There are currently only two services addressed in this manner. As a result it is practical to maintain a static mapping in a configuration file. Other methods have been used, including DNS.
 
 ## Path Construction (Segment Combinations) {#construction}
 
@@ -741,7 +737,7 @@ The one-hop path type `OneHopPath` is currently used to bootstrap beaconing betw
 
 A one-hop path has exactly one info field and two hop fields with the specialty that the second hop field is not known a priori, but is instead created by the ingress SCION border router of the neighboring AS while processing the one-hop path. Any entity with access to the forwarding key of the source endpoint AS can create a valid info and hop field as described in [](#inffield) and [](#hopfld), respectively.
 
-Upon receiving a packet containing a one-hop path, the ingress border router of the destination AS fills in the `ConsIngress` field in the second hop field of the one-hop path with the ingress interface ID. It sets the `ConsEgress` field to "0", indicating that the path cannot be used beyond the destination AS. Then it calculates and appends the appropriate MAC for the hop field.
+Upon receiving a packet containing a one-hop path, the ingress border router of the destination AS fills in the `ConsIngress` field in the second hop field of the one-hop path with the ingress interface ID. It sets the `ConsEgress` field to an invalid value, ensuring the path cannot be used beyond the destination AS. Then it calculates and appends the appropriate MAC for the hop field.
 
 {{figure-10}} below shows the layout of a SCION one-hop path type. There is only a single info field; the appropriate hop field can be processed by a border router based on the source and destination address. In this context, the following rules apply:
 
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -176,7 +176,7 @@ SCION leverages source-based path selection, where path information is embedded 
 
 **Data Plane**: The data plane (sometimes also referred to as the forwarding plane) is responsible for forwarding data packets that endpoints have injected into the network. After routing information has been disseminated by the control plane, packets are forwarded according to such information by the data plane.
 
-**Endpoint**: An endpoint is the start- or the endpoint of a SCION path. For example, an endpoint can be a host as defined in {{RFC1122}}, or a gateway bridging a SCION and an IP domain. This definition is based on the definition in {{RFC9473}}.
+**Endpoint**: An endpoint is the start- or the end of a SCION path. For example, an endpoint can be a host as defined in {{RFC1122}}, or a gateway bridging a SCION and an IP domain. This definition is based on the definition in {{RFC9473}}.
 
 **Forwarding Key**: A forwarding key is a symmetric key that is shared between the control service (control plane) and the routers (data plane) of an AS. It is used to authenticate hop fields in the end-to-end SCION path. The forwarding key is an AS-local secret and is not shared with other ASes.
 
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-dataplane.md
```diff
@@ -186,6 +186,8 @@ SCION leverages source-based path selection, where path information is embedded
 
 **Info Field (INF)**: Each path-segment construction beacon (PCB) contains a single info field, which provides basic information about the PCB. Together with hop fields (HFs), info fields are used to create forwarding paths.
 
+**Interface Identifier (Interface ID)**: A unique (unless 0) number within an AS that designates a network connection to another AS. Each interfaces of an AS is managed by one border router. Hop fields describe the traversal of an AS by a pair of interface IDs (the ingress and egress intefaces). Interface ID 0 denotes a connection between a border router and the AS's internal network. Each border router manages an instance of interface 0.
```
nicorusti, line 189:
```suggestion
**Interface Identifier (Interface ID)**: 16 bit identifier that designates a SCION interface at the end of a link connecting two SCION ASes. Each interface belongs to one border router. Hop fields describe the traversal of an AS by a pair of interface IDs (the ingress and egress interfaces). The Interface ID MUST be unique within each AS, with the exception of ID 0 that denotes a connection between a border router and the AS's internal network. Each border router manages an instance of interface 0.
```

draft-dekater-scion-dataplane.md
```diff
@@ -204,6 +206,8 @@ SCION leverages source-based path selection, where path information is embedded
 
 **Path Transparency**: Path transparency is a property of a network architecture that gives endpoints full visibility over the network paths their packets are taking. Path transparency is weaker than path control.
 
+**Peering Link**: A connection between two border routers, of different ASes and possibly different ISDs, that are not Core ASes. A peering link can be seen as a short-cut on a normal path which would otherwise include one or more core ASes. Peering link information is added to segment information during the beaconing process and used to shorten paths while assembling them from segments.
```
nicorusti, line 209: I'd use the same text I proposed for the CP
```suggestion
**Peering Link**: A link between two SCION border routers of different ASes, where at least one of the two ASes is not core. Two peering ASes may be in different ISDs. A peering link can be seen as a short-cut on a normal path. Peering link information is added to segment information during the beaconing process and used to shorten paths while assembling them from segments.
```

draft-dekater-scion-dataplane.md
```diff
@@ -250,6 +254,24 @@ The full forwarding process for a packet transiting an intermediate AS consists
...
+
+- Local interface ID
+- Neighbor type (core, parent, child, peer), depending on link type (see [](#paths-links)). Link type depends on mutual agreements between the organizations operating the ASes at each end of each link.
+- Neighbor ISD-AS number
+- Neighbor interface underlay address
+
+In addition, each router must be configured with a list of interface IDs in use within the AS, containing:
+
+- Interface ID
+- "intra-protocol" address of the remote router
```
nicorusti, line 269: Does that make sense? I tried to separate this in two lists, one for inter-AS neighbors and one for intra-AS SCION routers

draft-dekater-scion-dataplane.md
```diff
@@ -250,6 +254,24 @@ The full forwarding process for a packet transiting an intermediate AS consists
...
+
+In order to forward traffic to internal endpoints (traffic with interface ID 0), border routers need to associate endpoint addresses with concrete host addresses of the internal AS network. The nature of internal host addresses and the translation method are not specified by this document. They only depend on implementation and the choices of each AS's administrator. In current practice, internal host addresses are IP addresses (V6 or V4), both of which can be used as endpoint addresses in SCION headers; thereby requiring no translation.
```
nicorusti, line 271: Does this fit better in previous section  Intra-Domain Forwarding Process ?

draft-dekater-scion-dataplane.md
```diff
@@ -250,6 +254,24 @@ The full forwarding process for a packet transiting an intermediate AS consists
...
+
+In order to forward traffic to service endpoint addresses (`DT/DS` == 0b01) in the [common header](#common-header), a border router must translate service numbers into concrete host addresses. The method used to accomplish the translation is not defined by this document. It only depends on the implementation and the choices of each AS's administrator. In current practice this is accomplished by way of a configuration file. There are currently only two services addressed in this manner. As a result it is practical to maintain a static mapping in a configuration file. Other methods have been used, including DNS.
```
nicorusti, line 273: We don't describe this in this document.. This paragraph will be confusing. Let's discuss if this is needed. 

Also, we could just mention this mapping has to be configured or supplied out of band.

nicorusti (changes requested):

draft-dekater-scion-dataplane.md
jiceatscion: The phrase "intra-protocol" is completely obscure to me. We haven't used that previously, so we might want to make a longer, more explicit phrase. Otherwise, yeah that part probably deserves an explanation, but it might need to be longer.

draft-dekater-scion-dataplane.md
jiceatscion: I'd rather address the configuration topic in one place.

draft-dekater-scion-dataplane.md
jiceatscion: Well, we, unfortunately do mention the service address type, without explaining it. It is currently the only mechanism to bootstrap scion, so I think we are obligated to explain it.

draft-dekater-scion-dataplane.md
nicorusti: To be handled in #23

draft-dekater-scion-dataplane.md
```diff
@@ -186,6 +186,8 @@ SCION leverages source-based path selection, where path information is embedded
 
 **Info Field (INF)**: Each path-segment construction beacon (PCB) contains a single info field, which provides basic information about the PCB. Together with hop fields (HFs), info fields are used to create forwarding paths.
 
+**Interface Identifier (Interface ID)**: 16 bit identifier that designates a SCION interface at the end of a link connecting two SCION ASes. Each interface belongs to one border router. Hop fields describe the traversal of an AS by a pair of interface IDs (the ingress and egress interfaces). The Interface ID MUST be unique within each AS, with the exception of ID 0 that denotes a connection between a border router and the AS's internal network. Each border router manages an instance of interface 0.
```
nicorusti, line 189: We actually describe Interface ID 0 in the _One-Hop Path Type_  section. 
Maybe we could add a reference?   

Also, is interface ID 0 used for the last hop? In the example in _Step-by-Step Explanation_
```
      R4 will then also realize, based on the fields
      CurrHF and SegLen in the SCION header, that the packet has reached
      the last hop in its SCION path.  Therefore, instead of stepping up
      the pointers to the next info or hop field, router R4 inspects the
      SCION destination address and extracts the endpoint address
      192.0.2.7. 
 ```

File: draft-dekater-scion-dataplane.md (hunk 4 above)
Comment (jiceatscion) [reply draft-dekater-scion-dataplane.md]: Rearanged as per our discussion.

File: draft-dekater-scion-dataplane.md (hunk 6 above)
Comment (matzf) [reply draft-dekater-scion-dataplane.md:189]: > We actually describe Interface ID 0 in the One-Hop Path Type section.

The One hop path is not very special here; we also use 0 interface in the first/last hop field of normal SCION paths.

> Also, is interface ID 0 used for the last hop?

During path segment construction, the 0 interface is used in the first/last hop field. BUT in the path-segment-combined dataplane path, the 0 does not always/only occur on the last hop.
- Hops with 0 interface occurs "in the middle", at segment cross overs.
- Typically 0 will occur at the end of a down segment, but it's legal to cut off the end of a segment.
- Paths to a parent AS below the core consist of only an up-segment that is cut short. These don't end with 0.

File: draft-dekater-scion-dataplane.md [hunk 7]
```diff
@@ -246,9 +250,23 @@ The full forwarding process for a packet transiting an intermediate AS consists
 3. The SCION router maps the egress interface ID in the current hop field of the SCION header to the destination "intra-protocol" address of the egress border router (where "intra-protocol" is the intra-domain forwarding protocol, e.g., MPLS or IP).
 4. The packet is forwarded within the AS by routers and switches based on the "intra-protocol" header.
 5. Upon receiving the packet, the SCION egress router strips off the "intra-protocol" header, again validates and updates the SCION header, and forwards the packet to the neighboring SCION router.
+6. The last SCION router on the path forwards the packet to the packet's destination endpoint indicated by the field `DstHostAddr` of [the Address Header](#address-header).
+
+### Configuration
+
+Border routers require mappings from SCION  interface IDs to underlay addresses. Such information must be supplied to each router in an out of band fashion (e.g in a configuration file). For each link to a neighbor, these values must be configured:
+
+- Local interface ID
```
matzf, line 259: Just "Interface ID" (otherwise reader might wonder if there are different types of interface IDs).

draft-dekater-scion-dataplane.md
```diff
@@ -246,9 +250,23 @@ The full forwarding process for a packet transiting an intermediate AS consists
...
+- Neighbor type (core, parent, child, peer), depending on link type. Link type depends on mutual agreements between the organizations operating the ASes at each end of each link.
+- Neighbor ISD-AS number
+- Neighbor interface underlay address
+
+In addition, each router must be configured with a list of interface IDs in use within the AS, containing:
 
-**Note:** The current SCION implementation uses the UDP/IP protocol as underlay. However, the use of other underlay protocols is possible and allowed.
+- Interface ID
+- "intra-protocol" address of the remote router
```
matzf, line 267: Text does not make it quite clear which border routers get which list of values.

I'm guessing that it means to say that routers that "implement"/"handle" a particular interface need to know the link type/neighbor ISD-AS/neighbor interface address, while all other routers only need the address of the router that handles the Iface.

I believe all routers need to know the link type, to validate segment cross over (e.g. cross-over from an peer- to a core- link would be forbidden). Unfortunately, this is not described in the "processing at routers" section at all. Notably this _is_ described in the book v2 (5.6.3 processing at routers). 
The neighbor's ISD-AS number is currently not used for much, but possibly it should be used to drop packets with bad source/dest address.


Maybe it's easier to combine the two "lists". Suggestion:
```
For each link to a neighbor, these following information needs to be configured for all border routers:
- Interface ID
- Link Type,
- Neighbor ISD-AS number,
- In the case of the border router responsible for this interface: the neighbor interface's underlay address.
- Otherwise: the "intra-protocol" address of the border router responsible this interface.
```

draft-dekater-scion-dataplane.md
```diff
@@ -246,9 +250,23 @@ The full forwarding process for a packet transiting an intermediate AS consists
...
 
+In order to forward traffic to service endpoint addresses (`DT/DS` == 0b01) in the [common header](#common-header), a border router must translate service numbers into concrete host addresses. The method used to accomplish the translation is not defined by this document. It only depends on the implementation and the choices of each AS's administrator. In current practice this is accomplished by way of a configuration file. There are currently only two services addressed in this manner. As a result it is practical to maintain a static mapping in a configuration file. Other methods have been used, including DNS.
```
matzf, line 269: > (`DT/DS` == 0b01) in the [common header](#common-header)

"in the common header" should be in the parens, no?
-> (`DT/DS` == 0b01 in the common header) ...

draft-dekater-scion-dataplane.md
nicorusti: Thanks for the explanation! @jiceatscion can you remove mentions of this use of interface ID 0 from the draft?

draft-dekater-scion-dataplane.md
nicorusti: > Maybe it's easier to combine the two "lists"

But routers within the same AS require less parameters. For example, AFAIK no link type or neighbor ISD-AS is needed.  
Also, is it a requirement that all BRs within an AS see each others as "neighbors", in a meshed way? In other words, must each router have all interface ID mappings for its AS (and L3 connectivity to all other routers)?

draft-dekater-scion-dataplane.md
jiceatscion: yes. done.

draft-dekater-scion-dataplane.md
jiceatscion: These are implementation details, I would say. That section is an explanation about configuration; in response to a question. It is not actually mandating anything.

I rewote along the lines proposed my Matthias. We discuss that later.

draft-dekater-scion-dataplane.md
jiceatscion: ping?

draft-dekater-scion-dataplane.md
matzf: > But routers within the same AS require less parameters. For example, AFAIK no link type or neighbor ISD-AS is needed.

As I said above, they _do_ need to have the information about the link type.

>  Also, is it a requirement that all BRs within an AS see each others as "neighbors", in a meshed way? In other words, must each router have all interface ID mappings for its AS (and L3 connectivity to all other routers)?

Almost. The exact requirement is somewhat more nuanced. For the draft, I think this is good enough.

Just for the sake of completeness: the model of path segment crossover implicitly assumes that you can transit between any two child interfaces, and between any child and core interface. So, a router responsible for a child or core interface needs to know _at least_ all (other) child interfaces. All other hops (parent-child, core-core, child-peer), are announced explicitly; the corresponding connectivity only needs to exist if the hop is ever announced.

matzf (approved):

matzf (approved):

draft-dekater-scion-dataplane.md
```diff
@@ -723,7 +735,7 @@ The one-hop path type `OneHopPath` is currently used to bootstrap beaconing betw
 
 A one-hop path has exactly one info field and two hop fields with the specialty that the second hop field is not known a priori, but is instead created by the ingress SCION border router of the neighboring AS while processing the one-hop path. Any entity with access to the forwarding key of the source endpoint AS can create a valid info and hop field as described in [](#inffield) and [](#hopfld), respectively.
 
-Upon receiving a packet containing a one-hop path, the ingress border router of the destination AS fills in the `ConsIngress` field in the second hop field of the one-hop path with the ingress interface ID. It sets the `ConsEgress` field to "0", indicating that the path cannot be used beyond the destination AS. Then it calculates and appends the appropriate MAC for the hop field.
+Upon receiving a packet containing a one-hop path, the ingress border router of the destination AS fills in the `ConsIngress` field in the second hop field of the one-hop path with the ingress interface ID. It sets the `ConsEgress` field to an invalid value, ensuring the path cannot be used beyond the destination AS. Then it calculates and appends the appropriate MAC for the hop field.
```
nicorusti, line 738:
```suggestion
Upon receiving a packet containing a one-hop path, the ingress border router of the destination AS fills in the `ConsIngress` field in the second hop field of the one-hop path with the ingress interface ID. It sets the `ConsEgress` field to the unspecified value 0, ensuring the path cannot be used beyond the destination AS. Then it calculates and appends the appropriate MAC for the hop field.
```

How about we use the same wording as above, where we mention 0: 
```  *Interface Identifier (Interface ID)*: 16 bit identifier that
   designates a SCION interface at the end of a link connecting two
   SCION ASes.  Each interface belongs to one border router.  Hop fields
   describe the traversal of an AS by a pair of interface IDs (the
   ingress and egress interfaces).  The Interface ID MUST be unique
   within each AS.  Interface ID 0 is not a valid identifier,
   implementations can use it as the "unspecified" value.
   ```

nicorusti (approved):
LGTM

PR #26, opened 2024-06-04
Clarify selection of MAC, elaborate on MAC chaining and path splicing

matzf:
Describe AES-CMAC hop field MAC algorithm as the default which MUST be supported.

- Moved the details of  the default mac algorithm to _after_ the explanation of the accumulator field (as this is used in the mac input).
- Fixed the swapped info field/hop field labels in the figure depicting the MAC input.
- Describe requriements on alternative MAC implementations, in particular about requirements for MAC chaining
- Describe path splicing is prevented
- Describe how path splicing can still happen in security considerations

Contributes to https://github.com/scionassociation/scion-cp_I-D/issues/23.
Contributes to https://github.com/scionassociation/scion-dp_I-D/issues/7

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -1091,57 +1091,31 @@ When combining path segments to create a path to the destination endpoint, the s
 
 In the following, the computation of the hop field MAC as well as the accumulator field `Acc` is explained.
 
-**Note:** The algorithm used by SCION to compute the hop field MAC is based on the AES-CMAC algorithm, truncated to 48-bits - see also {{RFC4493}}. In principle, the computation of the MAC is an AS-specific choice; only the control service and routers of the AS need to agree on keys, algorithm, and input for the MAC. However, note that we do not provide nor specify any mechanism to coordinate AS-specific choices between the routers and the control services of the AS.
-
-
-#### MAC - Definition {#def-mac}
+#### Hop Field MAC - Definition {#def-mac}
 
 - Consider a path segment with "n" hops, containing ASes AS<sub>0</sub>, ... , AS<sub>n-1</sub>, with forwarding keys K<sub>0</sub>, ... , K<sub>n-1</sub> in this order.
 - AS<sub>0</sub> is the core AS that created the PCB representing the path segment and that added a random initial 16-bit segment identifier `SegID` to the `Segment Info` field of the PCB.
 
-The MAC<sub>i</sub> of the hop field of AS<sub>i</sub> is now calculated based on the following definition:
+The MAC<sub>i</sub> of the hop field of AS<sub>i</sub> is now calculated as:
 
 MAC<sub>i</sub> = <br> Ck<sub>i</sub> (`SegID` XOR MAC<sub>0</sub> \[:2] ... XOR MAC<sub>i-1</sub> \[:2], Timestamp, ExpTime<sub>i</sub>, ConsIngress<sub>i</sub>, ConsEgress<sub>i</sub>)
 
 where
 
 - k<sub>i</sub> = The forwarding key k of the current AS<sub>i</sub>
-- Ck<sub>i</sub> (...) = Checksum C over (...) computed with forwarding key k<sub>i</sub>
+- Ck<sub>i</sub> (...) = Cryptographic checksum C over (...) computed with forwarding key k<sub>i</sub>
 - `SegID` = The random initial 16-bit segment identifier set by the core AS when creating the corresponding PCB
 - XOR = The bitwise "exclusive or" operation
 - MAC<sub>i</sub> \[:2] = The hop field MAC for AS<sub>i</sub>, truncated to 2 bytes
 - Timestamp = The timestamp set by the core AS when creating the corresponding PCB
 - ExpTime<sub>i</sub>, ConsIngress<sub>i</sub>, ConsEgress<sub>i</sub> = The content of the hop field HF<sub>i</sub>
 
-The above definition implies that the current MAC is based on the XOR-sum of the truncated MACs of all preceding hop fields in the path segment as well as the path segment's `SegID`. In other words, the current MAC is *chained* to all preceding MACs.
-
-
-#### Layout of the Input Data for the MAC Calculation
-
-{{figure-17}} below shows the layout of the input data to calculate the hop field MAC in the data plane. The input layout represents the 8 bytes of the info field and the first 8 bytes of the hop field, with some fields zeroed out.
-
-Note that the MAC field itself is only 6 bytes long - see also [](#hopfld).
-
-~~~~
-0                   1                   2                   3
- 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ -----+
-|               0               |           Acc                 |   Hop|
-|-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-|-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-| Field|
-|                           Timestamp                           |      |
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-| -----+
-|       0       |    ExpTime    |          ConsIngress          |  Info|
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ Field|
-|          ConsEgress           |               0               |      |
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ -----+
-~~~~
-{: #figure-17 title="Input data to calculate the hop field MAC"}
-
+Thus, the current MAC is based on the XOR-sum of the truncated MACs of all preceding hop fields in the path segment as well as the path segment's `SegID`. In other words, the current MAC is *chained* to all preceding MACs.
 
 
 #### Accumulator Acc - Definition {#def-acc}
 
-The accumulator Acc<sub>i</sub> is an updatable counter introduced in the data plane to reduce overhead caused by MAC-chaining for path authorization. This is achieved by incrementally tracking the XOR-sum of the previous MACs as a separate, updatable accumulator field `Acc`, which is part of the path segment's info field `InfoField` in the packet header (see also [](#inffield)). Routers update this field by adding/subtracting only a single 16-bit value each.
+The accumulator Acc<sub>i</sub> is an updatable counter introduced in the data plane to avoid overhead caused by MAC-chaining for path authorization. This is achieved by incrementally tracking the XOR-sum of the previous MACs as a separate, updatable accumulator field `Acc`, which is part of the path segment's info field `InfoField` in the packet header (see also [](#inffield)). Routers update this field by adding/subtracting only a single 16-bit value each.
 
 ~~~~
  0                   1                   2                   3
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-dataplane.md
```diff
@@ -1091,57 +1091,31 @@ When combining path segments to create a path to the destination endpoint, the s
...
 
 
 #### Accumulator Acc - Definition {#def-acc}
 
-The accumulator Acc<sub>i</sub> is an updatable counter introduced in the data plane to reduce overhead caused by MAC-chaining for path authorization. This is achieved by incrementally tracking the XOR-sum of the previous MACs as a separate, updatable accumulator field `Acc`, which is part of the path segment's info field `InfoField` in the packet header (see also [](#inffield)). Routers update this field by adding/subtracting only a single 16-bit value each.
+The accumulator Acc<sub>i</sub> is an updatable counter introduced in the data plane to avoid overhead caused by MAC-chaining for path authorization. This is achieved by incrementally tracking the XOR-sum of the previous MACs as a separate, updatable accumulator field `Acc`, which is part of the path segment's info field `InfoField` in the packet header (see also [](#inffield)). Routers update this field by adding/subtracting only a single 16-bit value each.
```
jiceatscion, line 1118: > to avoid overhead caused by

to avoid the overhead caused by?

jiceatscion (approved):

draft-dekater-scion-dataplane.md
```diff
@@ -1175,6 +1149,31 @@ Acc<sub>i+1</sub> = Acc<sub>i</sub> XOR MAC<sub>i</sub> \[:2]
 - MAC<sub>i</sub> \[:2] = The hop field MAC for the current AS<sub>i</sub>, truncated to 2 bytes
 
 
+#### Default Hop Field MAC Algorithm
+
+The computation of the hop field MAC is an AS-specific choice. The operator of an AS can freely choose a MAC algorithm without outside coordination. However, the control service and routers of the AS do need to agree on the algorithm used.
+Control service and router implementations SHOULD support the Default Hop Field MAC algorithm described below. This document does not specify any further mechanism to coordinate this choice between control services and routers of one AS.
```
nicorusti, line 1155:
```suggestion
The algorithm used to compute the hop field MAC is an AS-specific choice. The operator of an AS can freely choose a MAC algorithm without outside coordination. However, the control service and routers of the AS MUST to agree on the algorithm used.
All control service and router implementations MUST support the Default Hop Field MAC algorithm described below. Operators may use other algorithms as long as they are supported by all devices in the AS.
```

Our reviewer Joel suggested use of MUST: 
> <jmh>Yes, I get that each AS can in principle use a different algorithm.  Now look at it from the point of view of an implementor.  What algorithm shall I implement?  I am expecting to sell to multiple operators.  I am expecting to need to interoperate with solutions from other vendors.  I have no problem with allowing a range of choices.  But I need something that ensures interoperability.  The usual answer is to say that all implementations must implement and support X, and that operators can use anything they like that their devices support as long as all devices in the AS use the same thing. </jmh>


Also, shoutln't only the algorithm be selectable (but not which fields ar MACed)? I propose a slightly different text to clarify this. I also did a bit of rewording based on Joel's comment

draft-dekater-scion-dataplane.md
```diff
@@ -1175,6 +1149,31 @@ Acc<sub>i+1</sub> = Acc<sub>i</sub> XOR MAC<sub>i</sub> \[:2]
...
+
+
+The default MAC algorithm is AES-CMAC ({{RFC4493}}) truncated to 48-bits, computed over the info field and the first 6 bytes of the hop field, with flags and reserved fields zeroed out. The input is padded to 16 bytes. The _first_ 6 bytes of the AES-CMAC output are used as resulting hop field MAC.
+{{figure-18}} below shows the layout of the input data to calculate the hop field MAC.
```
nicorusti, line 1159: I would clarify that this data must be used as an input to calculate MAC, regardless of algorithm

draft-dekater-scion-dataplane.md
```diff
@@ -1175,6 +1149,31 @@ Acc<sub>i+1</sub> = Acc<sub>i</sub> XOR MAC<sub>i</sub> \[:2]
 - MAC<sub>i</sub> \[:2] = The hop field MAC for the current AS<sub>i</sub>, truncated to 2 bytes
 
 
+#### Default Hop Field MAC Algorithm
+
+The computation of the hop field MAC is an AS-specific choice. The operator of an AS can freely choose a MAC algorithm without outside coordination. However, the control service and routers of the AS do need to agree on the algorithm used.
+Control service and router implementations SHOULD support the Default Hop Field MAC algorithm described below. This document does not specify any further mechanism to coordinate this choice between control services and routers of one AS.
+
+
+The default MAC algorithm is AES-CMAC ({{RFC4493}}) truncated to 48-bits, computed over the info field and the first 6 bytes of the hop field, with flags and reserved fields zeroed out. The input is padded to 16 bytes. The _first_ 6 bytes of the AES-CMAC output are used as resulting hop field MAC.
```
nicorusti, line 1158: Shall we mention that as MAC is truncated to 6 bits, in order to safely truncate it,  any other MAC algorithms must have properties of a pseudorandom function?

nicorusti (changes requested):

draft-dekater-scion-dataplane.md
```diff
@@ -1091,57 +1091,31 @@ When combining path segments to create a path to the destination endpoint, the s
...
-|               0               |           Acc                 |   Hop|
-|-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-|-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-| Field|
-|                           Timestamp                           |      |
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-| -----+
-|       0       |    ExpTime    |          ConsIngress          |  Info|
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ Field|
-|          ConsEgress           |               0               |      |
-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+ -----+
-~~~~
-{: #figure-17 title="Input data to calculate the hop field MAC"}
```
nicorusti, line 1138: Shoudln't the input data for MACs be always the same? To this point, isn't it better to leave this section where it was?

draft-dekater-scion-dataplane.md
matzf: I'll add a paragraph about the requirements for alternative MAC algos: unforgeable, pseudorandom in last two bytes (because of XOR thing), and depend on info field and hop field (ignoring flags).

This should constrain the input of the MAC function enough. I would avoid over-specifying the bit pattern layout of the input to alternative MAC algos; who knows, maybe it would make sense to separately compute something over some fields first, then some others, and then combine the output with some other function.

draft-dekater-scion-dataplane.md
matzf: The shortening to 6 bytes may be specific to the MAC algo. I think this should be covered by the more generic requirements that I alluded to in the comment above.

draft-dekater-scion-dataplane.md
matzf: See other comment.

draft-dekater-scion-dataplane.md
matzf: Could have sworn I fixed this. Thanks, done!

draft-dekater-scion-dataplane.md
nicorusti: > I would avoid over-specifying the bit pattern layout of the input to alternative MAC algos; who knows, maybe it would make sense to separately compute something over some fields first, then some others, and then combine the output with some other function.

I see your point. However, shoudln't we make sure that all these fields are in some way authenticated, so that the overall security properties are maintained? Otherwise you would allow an AS that only authenticates a portion of the fields. Maybe we could slightly reword it to say which fields must be authenticated?

draft-dekater-scion-dataplane.md
nicorusti: Ok thanks for the clarification

draft-dekater-scion-dataplane.md
nicorusti: Ok perfect, but ti would be good to clearly list which fields MUST be authenticated by the MAC.

matzf: done.

draft-dekater-scion-dataplane.md
matzf: Done.

draft-dekater-scion-dataplane.md
matzf: Done (applied most of your suggestion).

draft-dekater-scion-dataplane.md
```diff
@@ -1382,7 +1409,16 @@ In the unlikely case that an online brute-force attack succeeds, the obtained ho
...
+Finally, the hop field MAC protects the 16-bit aggregation of path segment identifier and preceding MACs. For details, see [](#auth-chained-macs). This MAC chaining prevents splicing even in the case that the AS interface and segment timestamp match.
+
+As the segment identifier and aggregation of preceding MACs is only 16-bits wide, per-chance collision among compatible path segments can occur.
+With typical network sizes and numbers of paths of today, such collisions might occur rarely.
+Successful path splicing access may result in a usable path that violates an ASes path policy, e.g. making a special transit link available to a customer AS that is not billed accordingly, or violate global validity as a spliced path segment could traverses one or multiple links twice.
+A wider security margin against path splicing could be obtained by increasing the width of the segment identifier / `Acc` field, e.g by extending it into the 8-bit reserved field next to it in the info field. Such a change in the protocol would have to be coordinated with all participating entities (control services, routers, end hosts in all ASes).
```
nicorusti, line 1421:
```suggestion
```
I'd not describe potential changes to the protocol, I don't think this is within the scope of the draft

nicorusti: @matzf let us know if you hear any feedback on this

draft-dekater-scion-dataplane.md
```diff
@@ -1089,6 +1102,8 @@ For high-speed packet processing in the **data plane**, computing even cheap ope
 
 When combining path segments to create a path to the destination endpoint, the source endpoint MUST also initialize the value of accumulator field `Acc` for each path segment. The `Acc` field MUST contain the correct XOR-sum of the path segment identifier and preceding hop field MACs expected by the first router that is traversed.
 
+The aggregated 16-bit path segment identifier and preceding MACs prevents the splicing parts of different path segments, unless there is a per-chance collision among compatible path segments in one AS. Path segments are "compatible" for splicing only if they enter the AS from the same ingress interface and have the same origination timestamp. With typical numbers of paths in
```
nicorusti, line 1141: >  With typical numbers of paths in 

Some part of the sentence is missing?

matzf: Ah good catch. I had started to write the analysis on path splicing here. Shortened, and replaced it with a reference to the security analysis.

draft-dekater-scion-dataplane.md
```diff
@@ -1431,7 +1458,16 @@ In the unlikely case that an online brute-force attack succeeds, the obtained ho
...
+Finally, the hop field MAC protects the 16-bit aggregation of path segment identifier and preceding MACs. For details, see [](#auth-chained-macs). This MAC chaining prevents splicing even in the case that the AS interface and segment timestamp match.
+
+As the segment identifier and aggregation of preceding MACs is only 16-bits wide, per-chance collision among compatible path segments can occur.
+With typical network sizes and numbers of paths of today, such collisions might occur rarely.
+Successful path splicing access may result in a usable path that violates an ASes path policy, e.g. making a special transit link available to a customer AS that is not billed accordingly, or violate global validity as a spliced path segment could traverses one or multiple links twice.
+A wider security margin against path splicing could be obtained by increasing the width of the segment identifier / `Acc` field, e.g by extending it into the 8-bit reserved field next to it in the info field. Such a change in the protocol would have to be coordinated with all participating entities (control services, routers, end hosts in all ASes).
```
nicorusti, line 1470:
```suggestion
Successful path splicing would allow an attacker to briefly use a path that violates an ASes path policy, e.g. making a special transit link available to a customer AS that is not billed accordingly, or violate global validity as a spliced path segment could traverses one or multiple links twice. Loops are excluded as they would require two simultaneous splices.
A wider security margin against path splicing could be obtained by increasing the width of the segment identifier / `Acc` field, e.g. by extending it into the 8-bit reserved field next to it in the info field.
```
Proposing an alternative text: 
* CLarify that the attacker would use the splice 
* Mention that loops would require two simultaneous splices
* Remove discussion about what is needed to change header. I also propose some text to describe loops.

nicorusti (approved):
Hi @matzf, I propose some text on path splicing, if that works for you, then we could finally merge this.. If not, let's briefly discuss Monday

draft-dekater-scion-dataplane.md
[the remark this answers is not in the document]
matzf: Done

matzf: Done

PR #27, opened 2024-06-13
remove allowed

### LINKED ISSUE #11: Allowed?
>  \*Note:\* The current SCION implementation uses the UDP/IP protocol as
   underlay.  However, the use of other underlay protocols is possible
   and allowed.

This last sentence makes it seem like one might never run SCION on its own.  Also, "allowed" isn't a word we should see in RFCs.

nicorusti:
Resolves #11

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -37,7 +37,6 @@ author:
 normative:
   RFC1122:
   RFC1918:
-  RFC2119:
   RFC2474:
   RFC2711:
   RFC3168:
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-dataplane.md
```diff
@@ -284,7 +284,7 @@ Segments cannot be combined arbitrarily. To construct a valid forwarding path, t
 
 Besides enabling the enforcement of path policies, the above rules also protect the economic interest of ASes, as they prevent building "valley paths". A valley path contains ASes that do not profit economically from traffic on this route. The name comes from the fact that such paths go "down" (following parent-child links) before going "up" (following child-parent links).
 
-{{figure-1}} below shows the different allowed segment combinations.
+{{figure-1}} below shows the possible segment combinations.
```
jiceatscion, line 287: "possible" doesn't really convey the effect of other choices. After all one can totall craft incorrect packets. How about "valid"? That's what the router does, after all: it "validates" the path.

draft-dekater-scion-dataplane.md
```diff
@@ -352,7 +352,7 @@ Besides enabling the enforcement of path policies, the above rules also protect
 {: #figure-1 title="Illustration of possible path-segment combinations. Each node represents a SCION Autonomous System."}
 
 
-The following path-segment combinations are allowed:
+Possible path-segment combinations:
```
jiceatscion, line 355: ditto

jiceatscion (approved):

jiceatscion (approved):

draft-dekater-scion-dataplane.md
nicorusti: Changed to Valid

nicorusti: Changed to Valid

Issue #23, opened 2024-05-30
intra-protocol? --> better wording 


nicorusti:
Some ideas:
- intra-domain underlay

Second choice:
- intra-domain network protocol

If we use a term, we may need to add it to the glossary too.

PR #29, opened 2024-07-04
Replaced 'intra-protocol' with 'intra-domain protocol' throughout

knmeynell:
Solves #23
I decided to use 'intra-domain protocol' rather than 'intra-domain network protocol' in the interests of conciseness, but also because intra-domain protocol was already defined in the document. Some further re-wording for improved readability and clarify.


nicorusti:
I had to also change this in the CP draft: https://github.com/scionassociation/scion-cp_I-D/pull/44

draft-dekater-scion-dataplane.md
```diff
@@ -1468,8 +1468,8 @@ In the unlikely case that an online brute-force attack succeeds, the obtained ho
 
 In a path-splicing attack, an adversary source endpoint takes valid hop fields of multiple path segments and splices them together to obtain a new unauthorized path.
 
-Two candidates path segments for splicing must have at least one AS interface in common as a connection point.
-The path segments must have the same origination timestamp, as this is directly protected by the hop field MAC. This can occur by chance, or if the two candidate path segments were originated as the same segment that then diverged and converged back.
+Two candidates path segments for splicing MUST have at least one AS interface in common as a connection point.
+The path segments MUST have the same origination timestamp, as this is directly protected by the hop field MAC. This can occur by chance, or if the two candidate path segments were originated as the same segment that then diverged and converged back.
```
nicorusti, line 1472:
```suggestion
Two candidates path segments for splicing must have at least one AS interface in common as a connection point.
The path segments must have the same origination timestamp, as this is directly protected by the hop field MAC. This can occur by chance, or if the two candidate path segments were originated as the same segment that then diverged and converged back.
```

draft-dekater-scion-dataplane.md
```diff
@@ -1485,7 +1485,7 @@ When an adversary sits on the path between the source and destination endpoint,
...
+An on-path adversary could modify the SCION path header, and replace the remaining part of path segments to the destination with different segments. Such replaced segments MUST include authorized segments, as otherwise the packet would be simply dropped on its way to the destination. The already traversed portion of the current segment and past segments can also be modified by the adversary (for instance, deleting and adding valid and invalid hop fields). On reply packets from the destination, the adversary can transparently revert the changes to the path header again. For instance, if an adversary M is an intermediate AS on the path of a packet from A to B, then M can replace the packet’s past path (leading up to, but not including M). The new path may not be a valid end-to-end path. However, when B reverses the path and sends a reply packet, that packet would go via M, which can then transparently change the invalid path back to the valid path to A. In addition, the endpoint address header can also be modified.
```
nicorusti, line 1488:
```suggestion
An on-path adversary could modify the SCION path header, and replace the remaining part of path segments to the destination with different segments. Such replaced segments must include authorized segments, as otherwise the packet would be simply dropped on its way to the destination. The already traversed portion of the current segment and past segments can also be modified by the adversary (for instance, deleting and adding valid and invalid hop fields). On reply packets from the destination, the adversary can transparently revert the changes to the path header again. For instance, if an adversary M is an intermediate AS on the path of a packet from A to B, then M can replace the packet’s past path (leading up to, but not including M). The new path may not be a valid end-to-end path. However, when B reverses the path and sends a reply packet, that packet would go via M, which can then transparently change the invalid path back to the valid path to A. In addition, the endpoint address header can also be modified.
```
We are atlaking about attacks, I'd not use keywords

PR #34, opened 2024-07-21
Proof of transit: mention why it is not needed

nicorusti:
Follow up to https://github.com/scionassociation/scion-cp_I-D/issues/52


draft-dekater-scion-dataplane.md
```diff
@@ -1490,9 +1490,7 @@ An on-path adversary could modify the SCION path header, and replace the remaini
...
-To prevent this attack and to defend against multiple on-path adversaries in general, proof of transit is required, which is not in scope for this document, given the marginal benefit.
+Moreover, packet integrity protection is not enough if there are two colluding adversaries on the path. These colluding adversaries can forward the packet between them using a different path than selected by the source endpoint: The first on-path attacker remodels the packet header arbitrarily, and the second on-path attacker changes the path back to the original source-selected path, such that the integrity check by the destination endpoint succeeds. Under the assumptions for this attack to be possible, each of the  malicious on-path adversary may already inspect/copy traffic, therefore this attack does not represent an additional security gap. To prevent this attack and to defend against multiple on-path adversaries in general, proof of transit is required, which is not in scope for this document, given the marginal benefit.
```
jiceatscion, line 1493: I propose a slightly different wording, because I think that, as it is, the last sentence seems to contradict the rest. May be the explanation below is clearer? Not binding, keep your text if you prefer it.

Moreover, packet integrity protection is not enough if there are two colluding adversaries on the path. These colluding adversaries can forward the packet between them using a different path than selected by the source endpoint: The first on-path attacker remodels the packet header arbitrarily, and the second on-path attacker changes the path back to the original source-selected path, such that the integrity check by the destination endpoint succeeds. However, such an attack is of little value. An on-path adversary can inspect/copy/disrupt the traffic that reaches it without diverting the traffic away from the sender-chosen path. For this reason proof-of-transit, which would be required to detect such an attack, has little benfit in the context of SCION and is not in scope for this document.

jiceatscion (approved):

draft-dekater-scion-dataplane.md
nicorusti: Thanks, I used a slightly edited version of your wording

Issue #56, opened 2024-08-25
Peering links clarifications


nicorusti:
We currently say that peering links are only possible if one of the two ASes is not core, but there is an open discussion. 

- [x] Clarify wether peering links between core ASes are possible following https://github.com/scionproto/scion/issues/4484
- [x] We might need to say more explicitly how a router can determine whether the current hop is a peering hop. We now only mention this in "Steps at Egress Border Router" and I feel text is not too clear: ` This would be the case if the current Hop Field is the *first* Hop Field of the segment, `

PR #60, opened 2024-08-29
Clarify the validity of peering links

jiceatscion:
In anticipation of the TC or Adrian confirming the irrelevance of the Core-Core peering prohibition, removed the associated language in the draft and explicitly state that Core-Core peering is possible. Also clarified that this still does not permit valley paths.

Also clarified the rules applied by a router to determine if a hop is peering hop.

Fixes to #56 

- [ ] ~If merged, it must be replicated across drafts~


nicorusti (approved):
LGTM, thanks a lot!

PR #62, opened 2024-09-10
layer2: rephrase the L2 AS-AS link description

### LINKED ISSUE #55: Life of a packet: is L2 link possible?
In 3. Life of a SCION Data Packet we have a L2 link between R1, R2. Is this something that comes from the book or is this actually possible today? Can any of the implementations actually support this?

I feel there are some missing pieces: 
* We'd need an [Assigned Protocol Number](https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml) that can be used by L2
* Is there any need for a L2 discovery mechanism?

@jiceatscion 

phrasing to be adjusted: 

```
*Note:* Although technically there is no need for a UDP/IP
      underlay if two routers are directly connected, the SCION
      implementation always uses a UDP/IP underlay in practice.  This is
      to enable a common interface for all routers.
  ```

jiceatscion:
Fixes #55


nicorusti (approved):
LGTM

PR #65, opened 2024-09-20
mtu: add clarifications

### LINKED ISSUE #39: Add discussion about MTU & fragmentation
> Feedback Harald Alvestrand
> The description of SCION refers to using UDP/IP as the internal substrate inside an AS, and as a framing protocol on inter-AS links. This means that the Internet MTU of 1500 bytes (1200 for IPv6) will be in effect.
Path hops take 12 bytes + a couple of “segment headers” of 16 bytes - which means that a 10-hop, 3-segment path will take 168 bytes for the routing information. This is enough to reduce the MTU significantly, and the MTU will be unpredictable and maybe variable, since the path for a given packet may change dynamically.
(The packet format, like UDP’s, permits 68K-byte packets. However, fragmentation is A Thing.)

Related issues: 
* CP draft: https://github.com/scionassociation/scion-cp_I-D/issues/57 
* Implementations: https://github.com/scionproto/scion/issues/2608 

Also: 
- [x] Add discussion about fragmentation
- [x] Adapt this text  in the Common Header  paragraph` The value of the traffic class bits in a received packet or fragment might differ from the value sent by the packet's source. `

jiceatscion:
Fixes #39


draft-dekater-scion-dataplane.md
```diff
@@ -1038,6 +1037,27 @@ This section explains what happens with the SCION packet header at each router,
...
+
+The MTU of each link may be discovered or administratively configured (current practice is for it to be configured). It must be less than or equal to the MTU of the link's underlay encapsulation or native link-layer in either direction.
+
+SCION assumes that the MTUs of a path segment remains correct for the life time of that segment. This is generally a safe assumption because:
+
+* Intra-AS network MTUs are a result of the network configuration of each AS and therefore predictable.
+
+* Inter-AS links MTU are normally under the joint control of the administrators of the two ASes involved and therefore equally predictable.
+
+Although that isn't the main use case, SCION allows inter-AS links to be routed through multiple IP routers. In that case, the link's MTU MUST be configured statically to a conservative value. 1280 is a safe value. The same approach applies to all cases where MTUs cannot be assumed to be stable.
```
nicorusti, line 1060:
```suggestion
```
I think this is an exceptional case that would confuse readers and it could be omitted

nicorusti (commented):
I proposed some changes. 

Also, would the # Fragmentation section be a better fit after the MTU section? I feel the MTU section has a lot of explanations that can provide more background for fragmentation

draft-dekater-scion-dataplane.md
jiceatscion: This isn't an exceptional case. It is used heavily by some deployments such as scion lab. I agree that this is not a desirable configuration where security is the objective, but this is an important feature for experimental deployments. I do not think we should hide it. In fact, I think we should mention it somewhere outside the limited context of MTU.

draft-dekater-scion-dataplane.md
```diff
@@ -1038,6 +1037,27 @@ This section explains what happens with the SCION packet header at each router,
 
 When destination Endpoint B wants to respond to source Endpoint A, it can just swap the source and destination addresses in the SCION header, reverse the SCION path, and set the pointers to the Info Fields and Hop Fields at the beginning of the reversed path (see also [](#reverse)).
 
+## Packet Fragmentation
+
+The UDP/SCION layer never fragments packets. Unlike IPV6, even the sending endpoint cannot fragment packets on behalf of applications. Applications need to comply with the MTU of the paths that they use.
+
+SCION is agnostic to datagram fragmentation by the underlay network layer (such as UDP/IP encapsulation). Implementations SHOULD allow MTU discovery to be enabled and fragmentation to be disabled (honoring the system defaults is normally sufficient). This is the RECOMMENDED setting. For inter-AS links, using a different configuration is the joint decision of the administrators of the two ASes involved. For intra-AS interfaces using a different configuration is the choice of that AS' administrator alone.
```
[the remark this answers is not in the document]
jiceatscion: I accepted your suggestion. I wouldn't bother with other protocols since they aren't specified anywhere (is there a TCP/SCION implementation?

jiceatscion:
> I proposed some changes.
> 
> Also, would the # Fragmentation section be a better fit after the MTU section? I feel the MTU section has a lot of explanations that can provide more background for fragmentation

Will take a look, will do if that doesn't cause more forward references.

draft-dekater-scion-dataplane.md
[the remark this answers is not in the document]
nicorusti: Not sure, maybe we should completely omit details about higher layer protocols and just use a general: 

> Higher layer protocols MUST comply with the MTU of the paths that they use.

draft-dekater-scion-dataplane.md
nicorusti: How about more generic phrasing: 
> Should the inter-AS link MTU be unpredictable (e.g. because the inter-AS link is deployed on top of an overlay), then the link's MTU MUST be configured statically to a conservative value. Experience shows that 1280 is a safe value.

jiceatscion: Sounds good. Done

draft-dekater-scion-dataplane.md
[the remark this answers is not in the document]
jiceatscion: Mmm. May be. Fragmentation is actually an IP feature, which UDP may or may not rely on. SCION doesn't have any fragmentation support. So, we probably should just say that and do away with the MUST/MUST not when it comes to UDP/SCION? That would implicitly settle the question for other transports.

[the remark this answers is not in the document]
nicorusti: I like it, short and clean. I committed it. Thank you!

draft-dekater-scion-dataplane.md
```diff
@@ -398,7 +399,7 @@ The SCION common header has the following packet format:
 {: #figure-3 title="The SCION common header packet format"}
 
 - `Version`: The version of the SCION common header. Currently, only version "0" is supported.
-- `TrafficClass` (`TraffCl` in the image above): The 8-bit long identifier of the packet's class or priority. The value of the traffic class bits in a received packet or fragment might differ from the value sent by the packet's source. The current use of the `TrafficClass` field for Differentiated Services and Explicit Congestion Notification is specified in {{RFC2474}} and {{RFC3168}}.
+- `TrafficClass` (`TraffCl` in the image above): The 8-bit long identifier of the packet's class or priority. The value of the traffic class bits in a received packet might differ from the value sent by the packet's source. The current use of the `TrafficClass` field for Differentiated Services and Explicit Congestion Notification is specified in {{RFC2474}} and {{RFC3168}}.
```
nicorusti, line 402: @jiceatscion Do ypu know why we mention fragment here? I think we should not. 
This text comes from https://github.com/scionproto/scion/blob/74878a477c7821fb66f8d6a39bdddced590bd91d/doc/protocols/scion-header.rst#L35

nicorusti (approved):
Besides the small doubt above, LGTM :) Thanks!

draft-dekater-scion-dataplane.md
nicorusti: OK, can you also change this in scionproto?

jiceatscion: Done. see https://github.com/scionproto/scion/pull/4635

PR #68, opened 2024-10-04
Expanded on SCION-IP Gateway

### LINKED ISSUE #40: Specify SCION-IP Gateways - outline
> Harald Alvestrand (meeting notes 26.07.2024)
Document SCION-IP gateways: If the goal is to document existing deployment, the gateways must be specified. This could be an additional ISE draft. The overall SCION draft “package” must have enough information to make a connection over scion

- [x] Prepare a description of the SIG (try in terms of functionality rather than how they work)
- [x] Send text to Eliot
- [x] Decide if this needs to be a separate draft (or if it can be a paragraph in one of the other drafts) --> No

The payload of a packet can be anything (does not affect SCION operation), but the header contains an integer saying which content it is, with the defined values being “UDP”, “TCP”, and some numbers for internal use. The process for constructing an IP packet from the SCION packet (or vice versa) is not described, but presumably involves mangling checksums - -dataplane describes a “virtual header” for constructing the checksum of things that require it, which implies that a gateway to non-SCION IP would mangle checksums inside the UDP or TCP packet.
(There are references in -overview to docs describing gateways. However, I can’t see them at the moment.)

>   My response: 
Yes, we don’t yet have a draft describing gateways in detail. There was [a presentation ](https://datatracker.ietf.org/meeting/118/materials/slides-118-panrg-operational-aspects-of-scion-00)by my co-author Sam Hitz at PANRG at 119 about it. Essentially gateways run an encapsulation protocol and exchange IP routes among each other. They use a custom protocol for that, so we felt for now it was best to leave this part out and treat gateways as a normal SCION endpoint. 
This checksums is used, for example, by UDP on top of SCION. Gateways encapsulate the whole IP packet, so they don’t mangle UDP/TCP checksums in them. However, the gateway encapsulation protocol itself runs on top of UDP on top of SCION, therefore there is a ckechsum in the encapsulating UDP packet, and this uses the “virtual header”.

> Lunch discussion
Document SCION-IP gateways: If the goal is to document existing deployment, the gateways must be specified. This could be an additional ISE draft. The overall SCION draft “package”. The  must have enough information to make a connection over scion

#### Potential sources: 
- [Anapaya Docs](https://docs.anapaya.net/en/latest/edge/configuration/scion-tunneling/#edge-tunneling-sgrp) for SGRP, tunneling, also called "Prefix discovery" in open source
- [SCIONproto](https://docs.scion.org/en/latest/sig.html) for SIG framing 
- Scion book section  13.3 The SCION–IP Gateway (SIG) 
- [component analysis draft](https://www.ietf.org/archive/id/draft-rustignoli-panrg-scion-components-03.txt) section 3.1
- https://github.com/scionproto/scion/issues/2948

knmeynell:
Resolves #40

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -1872,7 +1872,9 @@ All fields within SIG Frame Header are in network byte order.
 - `Index` (16 bits) is the byte offset of the first beginning of an IP packet within the payload. If no IP packet starts in the payload, e.g. if the frame contains only the middle or trailing part of an IP packet, the field MUST be set to 0xFFFF.
 - `Reserved` (12 bits): it MUST be set to zero.
 - `Stream ID` (20 bits), along with the session, it identifies a unique sequence of SIG frames. Frames from the same stream are, on the remote SIG, put into the same reassembly queue. There may be multiple streams per session.
-- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames can be used to reassemble IP packets split among multiple frames. In the rare event that a frame is delayed or lost and a frame other than the next expected one is received, any previously received frames pertaining to a yet incomplete IP packet, are discarded. Subsequent frames are discarded until one carries the beginning of an IP packet. Late frames (with a sequence number older than any non-discarded ones) are discarded as they necessarily belong to an already discarded IP packet."
+- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames are used to reassemble IP packets split among multiple frames.
+
+SIGs MAY reorder received frames within a stream. Choice of frame reassembly window is left to implementations.
 
 The Session ID and Stream ID are chosen by the sender but the tuple MUST be unique within a session.
 
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -1805,7 +1805,7 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
 
 # Interoperability: SCION IP Gateway {#sig}
 
-The SCION IP Gateway (SIG) enables IP packets to be tunnelled over SCION to support communication between hosts that do not run a SCION implementation. A SIG acts as a router from the perspective of IP, whilst acting as SCION endpoint from the perspective of the SCION network. It is typically deployed inside the same AS-internal network as its non-SCION hosts, or at the edge of an enterprise network.
+The SCION IP Gateway (SIG) enables IP packets to be tunneled over SCION to support communication between hosts that do not run a SCION implementation. A SIG acts as a router from the perspective of IP, whilst acting as SCION endpoint from the perspective of the SCION network. It is typically deployed inside the same AS-internal network as its non-SCION hosts, or at the edge of an enterprise network.
 
 Tunneling IP traffic over SCION requires a pair of SIGs and it involves the following steps:
 
@@ -1813,21 +1813,20 @@ Tunneling IP traffic over SCION requires a pair of SIGs and it involves the foll
 
 2. The IP packet reaches a SIG in the sender’s network via standard IP routing.
 
-3. Based on the destination IP address, the source SIG determines the destination SIG's ISD-AS, endpoint address. To achieve this, SIGs may be pre-configured with static IP prefix to remote SIG SCION address mappings. Alternatively, a pair of SIGs may be configured to speak a dynamic routing protocol between each other. Choice of protocol is outside of the scope of this document and it is left up to implementors.
+3. Based on the destination IP address, the source SIG determines the destination SIG's ISD-AS endpoint address. To achieve this, SIGs may be pre-configured with a static IP prefix to remote SIG SCION address mappings. Alternatively, a pair of SIGs may be configured to speak a dynamic routing protocol between each other. The choice of protocol is left up to implementors and is outside of the scope of this document.
 
-4. The SIG encapsulates the original IP packet within a SCION packet and sends it to the remote SIG. If necessary, the source SIG performs a SCION path lookups and selects a SCION path to the destination SIG.
+4. The SIG encapsulates the original IP packet within a SCION packet and sends it to the remote SIG. If necessary, the source SIG performs SCION path lookups and selects a SCION path to the destination SIG.
 
-5. The remote SIG receives the SCION packet and decapsulates the original IP packet. It then forwards the packet to the final IP destination using standard local IP routing.
+5. The remote SIG receives the SCION packet and decapsulates the original IP packet. It then forwards the packet to the final IP destination using standard IP routing.
 
 ## SIG Framing
-IP packets are encapsulated over SCION/UDP into SIG frames. While in principle, a pair of SIGs may use other tunneling protocols, existing deployments leverage SIG framing as described here.
-TODO: @oncilla @shitz could you say something about why this framing and not existing protocol? Is it because of MTU?
+
+IP packets are encapsulated over SCION/UDP into SIG frames. Whilst in principle, a pair of SIGs may use other tunneling protocols, existing deployments use SIG framing as described here. This is to provide independence from the SCION MTU which can increase and decrease over time, to provide fast detection of packet loss and subsequent recovery of decapsulation for packets that weren't lost, and support for multiple streams within a framing session such that the streams can be distributed to separate cores.
 
 There may be multiple IP packets in a single SIG frame, and a single IP packet may be split into multiple SIG frames.
-A pair of SIGs establishes a SIG tunneling session.
-TODO: how are the sessions established? Is a session mapped to a single path?
-Within each session there may be multiple streams which is useful to distinguish between traffic sent by different SIG instances. For example, if an ingress SIG is restarted, it will create a new Stream ID for each session so the egress SIG will know that the new frame with a new Stream ID does not carry trailing part of the unfinished IP packet from a different stream.
-TODO: text above here seems contradicting. If a SIG is restarted, shouldn't it use a new session ID?
+A pair of SIGs establishes a SIG tunneling session which are is unidirectionally established by the sender. Whether a session is mapped to a single path or not is an implementation decision.
+
+Streams can be used within a session to distinguish between different packet flows. A single stream has a single reassembly queue and should be processed by a single core. Packets in a stream should not be distributed over multiple paths because that will lead to reordering and performance will drop.
 
 Each SIG frame has a sequence number that is used by the egress SIG to reassemble the encapsulated IP packets within a stream.
 
@@ -1873,11 +1872,10 @@ All fields within SIG Frame Header are in network byte order.
 - `Index` (16 bits) is the byte offset of the first beginning of an IP packet within the payload. If no IP packet starts in the payload, e.g. if the frame contains only the middle or trailing part of an IP packet, the field MUST be set to 0xFFFF.
 - `Reserved` (12 bits): it MUST be set to zero.
 - `Stream ID` (20 bits), along with the session, it identifies a unique sequence of SIG frames. Frames from the same stream are, on the remote SIG, put into the same reassembly queue. There may be multiple streams per session.
-- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames can be used to reassemble IP packets split among multiple frames.
+- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames can be used to reassemble IP packets split among multiple frames. In the rare event that a frame is delayed or lost and a frame other than the next expected one is received, any previously received frames pertaining to a yet incomplete IP packet, are discarded. Subsequent frames are discarded until one carries the beginning of an IP packet. Late frames (with a sequence number older than any non-discarded ones) are discarded as they necessarily belong to an already discarded IP packet."
 
-TODO:
-- how is the session established and torn down?
-- How is the stream ID set randomly? Is it negotiated? Is it associated with a session? Does it have to be unique within a session? Is it unidirectional?
+The Session ID and Stream ID are chosen by the sender but the tuple MUST be unique within a session.
+
 
 ## SIG Frame Payload
 
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -1805,23 +1805,29 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
 
 # Interoperability: SCION IP Gateway {#sig}
 
-The SCION IP Gateway (SIG) enables IP packets to be tunnelled over SCION to support the use of applications that are not SCION enabled.
+The SCION IP Gateway (SIG) enables IP packets to be tunnelled over SCION to support communication between hosts that do not run a SCION implementation. A SIG acts as a router from the perspective of IP, whilst acting as SCION endpoint from the perspective of the SCION network. It is typically deployed inside the same AS-internal network as its non-SCION hosts, or at the edge an enterprise network.
 
-An ingress SIG encapsulates IP packets within SCION packets and sends them across a SCION network to an egress SIG. The egress SIG decapsulates the IP packets from the SCION packets and forwards them towards their destination IP address. The SIGs at either end of a tunnel act as routers from the perspective of IP, whilst acting as SCION endpoints from the perspective of the SCION network.
+Tunneling IP traffic over SCION requires a pair of SIGs and it involves the following steps:
 
-An ingress SIG may choose to send SCION packets to an egress SIG in accordance with static IP routes, or by dynamically announcing IP prefixes to each other via a routing protocol. The ingress SIG is responsible for path selection to the egress SIG, and may also choose how to send SCION packets based on locally configured policies when multiple egress SIGs are available.
+1. A sender sends an IP packet towards an IP destination.
 
-A SIG is typically deployed inside the same AS internal network as its non-SCION hosts, or at the edge an enterprise network.
+2. The IP packet reaches a SIG in the sender’s network via standard IP routing.
+
+3. Based on the destination IP address, the source SIG determines the destination SIG's ISD-AS, endpoint address. To achieve this, SIGs may be pre-configured with static IP prefix to remote SIG SCION address mappings or they may speak a dynamic routing protocol with the corresponding remote SIG.
+
+4. The SIG encapsulates the original IP packet within a SCION packet and sends it to the remote SIG. If necessary, the source SIG performs a SCION path lookups and selects a SCION path to the destination SIG.
+
+5. The remote SIG receives the SCION packet and decapsulates the original IP packet. It then forwards the packet to the final IP destination using standard local IP routing.
 
 ## SIG Framing
 
-Each ingress SIG establishes a session to one or multiple egress SIGs and IP packets are encapsulated in SIG frames. There can be multiple IP packets in a single SIG frame, and a single IP packet can also be split into multiple SIG frames.
+IP packets are encapsulated over SCION/UDP into SIG frames. There may be multiple IP packets in a single SIG frame, and a single IP packet may be split into multiple SIG frames.
 
-SIG traffic MAY be sent over multiple SIG sessions, which MAY be used transport different classes of traffic (e.g. priority vs. normal.). Within each session there may be multiple streams which is useful to distinguish between traffic sent by different SIG instances. For example, if an ingress SIG is restarted, it will create a new Stream ID for each session so the egress SIG will know that the new frame with a new Stream ID does not carry trailing part of the unfinished IP packet from a different stream.
+A pair of SIGs establishes a SIG tunneling session. TODO: how?.  Within each session there may be multiple streams which is useful to distinguish between traffic sent by different SIG instances. For example, if an ingress SIG is restarted, it will create a new Stream ID for each session so the egress SIG will know that the new frame with a new Stream ID does not carry trailing part of the unfinished IP packet from a different stream.
 
-Each SIG frame has a sequence number is used by the egress SIG to reassemble the encapsulated IP packets.
+Each SIG frame has a sequence number is used by the egress SIG to reassemble the encapsulated IP packets within a stream.
 
-SIG framing within a SCION packet:
+~~~~
 
 +-----------------------+
 |         SCION         |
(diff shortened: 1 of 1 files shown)


nicorusti (changes requested):
I made a few changes: 

- step by step description of steps involved when forwarding traffic between two SIGs (based on https://docs.anapaya.net/en/latest/edge/configuration/scion-tunneling/#overview) 
- Fixing figures
- Added some TODOs

- [ ] @oncilla we felt we had to add more to address Harald and Eliot's concerns. Can you have an overall look at this section, and let us know if this is ok?
- [ ] @jiceatscion can you help clarifying the frame fields? (e.g. correctness of field lengths, descriptions)
- [ ] How are some fields set? How about sessions?
- [ ] Also, shall we move the SIG section before security considerations? I think that section is always at the end

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,82 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
 
 # Interoperability: SCION IP Gateway {#sig}
 
-The SCION IP Gateway (SIG) enables IP packets to be tunnelled over SCION to support the use of applications that are not SCION enabled.
+The SCION IP Gateway (SIG) enables IP packets to be tunnelled over SCION to support communication between hosts that do not run a SCION implementation. A SIG acts as a router from the perspective of IP, whilst acting as SCION endpoint from the perspective of the SCION network. It is typically deployed inside the same AS-internal network as its non-SCION hosts, or at the edge an enterprise network.
```
jiceatscion, line 1808: "at the edge an enterprise network" -> "at the edge of an enterprise network"

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,82 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+Tunneling IP traffic over SCION requires a pair of SIGs and it involves the following steps:
 
-Each SIG establishes a session to one or multiple remote SIGs. Each pair of SIGs uses a common tunneling protocol. Each SIG may choose to send SCION packets to a remote SIG in accordance with static IP routes, or by dynamically announcing IP prefixes to each other via a routing protocol. Each SIG may also choose how to send SCION packets based on locally configured policies when multiple remote SIGs are available.
-In addition, the source SIG is responsible for SCION path selection.
+1. A sender sends an IP packet towards an IP destination.
+
+2. The IP packet reaches a SIG in the sender’s network via standard IP routing.
+
+3. Based on the destination IP address, the source SIG determines the destination SIG's ISD-AS, endpoint address. To achieve this, SIGs may be pre-configured with static IP prefix to remote SIG SCION address mappings or they may speak a dynamic routing protocol with the corresponding remote SIG.
```
jiceatscion, line 1816: "with the corresponding remote SIG" -> "with other SIGs". There cannot be a *corresponding* remote SIG until it has been found by way of said dynamic routing protocol or configuration (the latter removing the need to speak for that purpose).

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,82 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+
+4. The SIG encapsulates the original IP packet within a SCION packet and sends it to the remote SIG. If necessary, the source SIG performs a SCION path lookups and selects a SCION path to the destination SIG.
+
+5. The remote SIG receives the SCION packet and decapsulates the original IP packet. It then forwards the packet to the final IP destination using standard local IP routing.
+
+## SIG Framing
+
+IP packets are encapsulated over SCION/UDP into SIG frames. There may be multiple IP packets in a single SIG frame, and a single IP packet may be split into multiple SIG frames.
+
+A pair of SIGs establishes a SIG tunneling session. TODO: how?.  Within each session there may be multiple streams which is useful to distinguish between traffic sent by different SIG instances. For example, if an ingress SIG is restarted, it will create a new Stream ID for each session so the egress SIG will know that the new frame with a new Stream ID does not carry trailing part of the unfinished IP packet from a different stream.
```
jiceatscion, line 1826: "distinguish between traffic sent by different SIG instances" -> does this mean "different SIG /incarnations/ ?"... that's what the following example seems to imply.

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,82 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+
+Each SIG frame has a sequence number is used by the egress SIG to reassemble the encapsulated IP packets within a stream.
```
jiceatscion, line 1828: "has a sequence number is used" -> "has a sequence number that is used"

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,82 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+
+- `Version` (8 bits) indicates the SIG framing version. It MUST be set to zero.
+- `Session` (8 bits) identifies a tunneling session between a pair of SIGs.
+- `Index` (16 bits) is the byte offset of the first beginning of an IP packet within the payload. If no IP packet starts in the payload, e.g. if the frame contains only a trailing part of an IP packet, the field MUST be set to 0xFFFF.
+- `Reserved` (12 bits): it MUST be set to zero.
+- `Stream` (20 bits), along with the session, it identifies a unique sequence of SIG frames. Frames from the same stream are, on the remote SIG, put into the same reassembly queue. There may be multiple streams per session.
+- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames can be used to reassemble IP packets split among multiple frames.
+
+TODO:
+- how is the session established and teared down?
```
jiceatscion, line 1875: Me no speak English good, but... "established and teared down?": isn't that supposed to be "established and torn down"?

draft-dekater-scion-dataplane.md
knmeynell: Okay

draft-dekater-scion-dataplane.md
knmeynell: Okay

draft-dekater-scion-dataplane.md
knmeynell: 'established and torn down' would be correct.

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,82 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+
+~~~~
+{: #figure-29 title="SIG Frame Header format"}
+
+
+All fields within SIG Frame Header are in network byte order.
+
+- `Version` (8 bits) indicates the SIG framing version. It MUST be set to zero.
+- `Session` (8 bits) identifies a tunneling session between a pair of SIGs.
+- `Index` (16 bits) is the byte offset of the first beginning of an IP packet within the payload. If no IP packet starts in the payload, e.g. if the frame contains only a trailing part of an IP packet, the field MUST be set to 0xFFFF.
```
jiceatscion, line 1869: "if the frame contains only a trailing part of an IP packet": I think it would be more accurate to say "if the frame contains only the middle part of an IP packet". It is more likely than the trailing part, so a more relevant example". The case where it is the trailing part is a corner case; the packet would have to end exactly where the frame ends for there to be no beginning of packet there. That must happen a bit more often than chance alone, due to the fact that we can't fit a fragment shorter than 20 bytes at the end of a frame, but still a minority case, I think.

draft-dekater-scion-dataplane.md
jiceatscion, line 1826: I admit that I do not know exactly when/why new sessions are established. It is probably an important and not part of the wire format part of the protocol (so-called choreography) for inter-SIG compatibility. Should we attempt at describing it. I can research it in the code.

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,82 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+- `Reserved` (12 bits): it MUST be set to zero.
+- `Stream` (20 bits), along with the session, it identifies a unique sequence of SIG frames. Frames from the same stream are, on the remote SIG, put into the same reassembly queue. There may be multiple streams per session.
+- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames can be used to reassemble IP packets split among multiple frames.
```
jiceatscion, line 1872: Should we say what happens if a frame goes missing/re-ordered and the sequence cannot be re-assembled? The code we have in the OS impl is pretty rustic. It makes the optimistic assumption that this rarely happens and discards everything in its buffer when a gap appears (i.e. when a frame arrives that is not the next one in sequence). So just like IP fragmentation: fragment re-ordering is the death of the packet.

jiceatscion (commented):
Made some suggestions.

draft-dekater-scion-dataplane.md
nicorusti: Yes, can you find something more about how these sessions are established?

draft-dekater-scion-dataplane.md
nicorusti: @knmeynell I did not see this change in the text,  committed it now.

draft-dekater-scion-dataplane.md
nicorusti: Thanks for catching i. It is a TODO anyways, where can we find something more  about this?

draft-dekater-scion-dataplane.md
nicorusti: How about: 
> e.g. if the frame contains only the middle or trailing part of an IP packet

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,82 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+{: #figure-29 title="SIG Frame Header format"}
+
+
+All fields within SIG Frame Header are in network byte order.
+
+- `Version` (8 bits) indicates the SIG framing version. It MUST be set to zero.
+- `Session` (8 bits) identifies a tunneling session between a pair of SIGs.
+- `Index` (16 bits) is the byte offset of the first beginning of an IP packet within the payload. If no IP packet starts in the payload, e.g. if the frame contains only the middle or trailing part of an IP packet, the field MUST be set to 0xFFFF.
+- `Reserved` (12 bits): it MUST be set to zero.
+- `Stream` (20 bits), along with the session, it identifies a unique sequence of SIG frames. Frames from the same stream are, on the remote SIG, put into the same reassembly queue. There may be multiple streams per session.
```
nicorusti, line 1871: Is this set randomly or not?  Here comment says random, but in practice it is an increasing number
https://github.com/scionproto/scion/blob/aa917bb458d6675efc55b4a23fda2c694094553e/gateway/dataplane/encoder.go#L196

draft-dekater-scion-dataplane.md
nicorusti: absolutely, maybe this can be a couple of sentences about fragmentation? 
I'm not sure if we can write about a potentially better approach?

jiceatscion: I would go as far as embedding a design doc in the draft.

I would add the following to the `Sequence number` paragraph:
"Frames can, rarely, be delayed or lost. When a frame other than the next expected one is received, any previously received frames, pertaining to a yet incomplete IP packet, are discarded. Subsequent frames are discarded until one carries the beginning of an IP packet. Late frames (with a sequence number older than any non-discarded ones) are discarded as they necessarily belong to an already discarded IP packet."

draft-dekater-scion-dataplane.md
jiceatscion: Yeah, that works too :-)

draft-dekater-scion-dataplane.md
nicorusti: Could we say more generally that the SIG should take care of packer reordering but that further details are left to implementors?

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,86 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+
+3. Based on the destination IP address, the source SIG determines the destination SIG's ISD-AS, endpoint address. To achieve this, SIGs may be pre-configured with static IP prefix to remote SIG SCION address mappings. Alternatively, a pair of SIGs may be configured to speak a dynamic routing protocol between each other. Choice of protocol is outside of the scope of this document and it is left up to implementors.
+
+4. The SIG encapsulates the original IP packet within a SCION packet and sends it to the remote SIG. If necessary, the source SIG performs a SCION path lookups and selects a SCION path to the destination SIG.
+
+5. The remote SIG receives the SCION packet and decapsulates the original IP packet. It then forwards the packet to the final IP destination using standard local IP routing.
+
+## SIG Framing
+IP packets are encapsulated over SCION/UDP into SIG frames. While in principle, a pair of SIGs may use other tunneling protocols, existing deployments leverage SIG framing as described here.
+TODO: @oncilla @shitz could you say something about why this framing and not existing protocol? Is it because of MTU?
```
nicorusti, line 1824: @oncilla @shitz do you have some background on why SIGs use their own framing protocol rather than any of the existing ones?
@oncilla I remember you mentioned at some point that no other protocol fulfilled the needs. Which needs? Is it because SIG framing can fragment an IP packet in multiple frames, without incurring into MTU problems?

shitz: I cannot remember what lead to the decision to create a custom framing protocol. There are certainly some requirements/properties that we wanted, e.g., non-reliable transport, independence of SCION MTU (because that can increase and decrease over time), fast detection of packet loss and subsequent recovery of decapsulation for packets that weren't lost, support for multiple streams within a framing session such that the streams can be distributed to separate cores.
I'm not sure if there was an investigation whether an existing framing protocol would have all of these properties or if it simply wasn't done, but at least there is some justification why the framing protocol is the way it is.

draft-dekater-scion-dataplane.md
jiceatscion: My notes, before I start forgetting... the code is tough read:

* Sessions are unidirectional and initiated by senders upon discovering other gateways.
* Gateway are discovered only in remote IAs that are statically configured.
* The initiator gateway allocates one session per (receiverGateway, sessionPolicy) pair.
* All initiated sessions have a unique ID that is simply incremented from 0. The numbering is re-done every time the set of remote gateways changes. To keep the numbers ~stable, the sessions are ordered by (policy>IA). I haven't searched for the effects of session re-IDing.
* Since there are only 256 session IDs they are not necessarily unique, but they are unique within a given (initiator, receiver) pair, as long as there aren't 256+ session policies for a given remote IA.
* From a given initiator to a given receiver, the traffic of all endpoint pairs is multiplexed on the same session.
* All paths between two gateways can be used by any given one session.
* All paths, regardless of session get their own sender, queue, and ~random streamID.
* The receiver side just follows these. It will accept any connection (I haven't checked if sending IAs are checked against the static config.
* The receiver side has one worker for each incoming session, that is each (sender gateway, sessionID) pair. It is created on the fly as needed.
* Each worker keeps a different reassembly queue for each streamID (but streams share the same worker task, unlike on the sending side).

Conclusion, regarding sessionID and StreamID. Each (sessionID, streamID) pair gets its own re-assembly queue. The pairs are ~unique within the context where it matters (one gateway->gateway pair) by virtue of sessionID being unique, but only approximately: there can be streamID collisions with P=O(10^-9) as long as there aren't too many paths. It might hurt a bit when that happens.

draft-dekater-scion-dataplane.md
shitz: This is completely implementation specific. Just because the open-source implementation has a reassembly window of a single fragment doesn't mean that you could easily have implementations that for exmaple allow for up to N out-of-order fragments because they keep them in a buffer and reassemble the packet once all fragments have been received. I would not put this in the draft and state that it's implementation specific. This is even a completely local decision for each SIG.

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,86 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+
+- `Version` (8 bits) indicates the SIG framing version. It MUST be set to zero if following this specification.
+- `Session ID` (8 bits) identifies a tunneling session between a pair of SIGs.
+- `Index` (16 bits) is the byte offset of the first beginning of an IP packet within the payload. If no IP packet starts in the payload, e.g. if the frame contains only the middle or trailing part of an IP packet, the field MUST be set to 0xFFFF.
+- `Reserved` (12 bits): it MUST be set to zero.
+- `Stream ID` (20 bits), along with the session, it identifies a unique sequence of SIG frames. Frames from the same stream are, on the remote SIG, put into the same reassembly queue. There may be multiple streams per session.
+- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames can be used to reassemble IP packets split among multiple frames.
+
+TODO:
+- how is the session established and torn down?
```
shitz, line 1879: > how is the session established and torn down?
As stated above, there is no real session establishment. The sender chooses session ID, stream ID unilateraly. 

> How is the stream ID set randomly? Is it negotiated? Is it associated with a session? Does it have to be unique within a session? Is it unidirectional?

1. Chosen unilaterally by the sender however it sees fit
2. no
3. (session id, stream id) tuple needs to be unique
4. yes
5. yes

shitz (changes requested):

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,84 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+
+4. The SIG encapsulates the original IP packet within a SCION packet and sends it to the remote SIG. If necessary, the source SIG performs SCION path lookups and selects a SCION path to the destination SIG.
+
+5. The remote SIG receives the SCION packet and decapsulates the original IP packet. It then forwards the packet to the final IP destination using standard IP routing.
+
+## SIG Framing
+
+IP packets are encapsulated over SCION/UDP into SIG frames. Whilst in principle, a pair of SIGs may use other tunneling protocols, existing deployments use SIG framing as described here. This is to provide independence from the SCION MTU which can increase and decrease over time, to provide fast detection of packet loss and subsequent recovery of decapsulation for packets that weren't lost, and support for multiple streams within a framing session such that the streams can be distributed to separate cores.
```
nicorusti, line 1824:
```suggestion
IP packets are encapsulated over SCION/UDP into SIG frames. Whilst in principle, a pair of SIGs may use other tunneling protocols, existing deployments use SIG framing as described here. This is to provide independence from the underlying SCION path MTU which can increase and decrease over time, to provide fast detection of packet loss and subsequent recovery of decapsulation for packets that weren't lost, and support for multiple streams within a framing session such that the streams can be distributed to separate cores.
```

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,84 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+
+There may be multiple IP packets in a single SIG frame, and a single IP packet may be split into multiple SIG frames.
+A pair of SIGs establishes a SIG tunneling session which are is unidirectionally established by the sender. Whether a session is mapped to a single path or not is an implementation decision.
```
nicorusti, line 1827:
```suggestion
A source SIG unidirectionally establishes a SIG tunneling session with a destination SIG. A session may be mapped to one or multiple SCION paths, depending on implementors.
```

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,84 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+
+Streams can be used within a session to distinguish between different packet flows. A single stream has a single reassembly queue and should be processed by a single core. Packets in a stream should not be distributed over multiple paths because that will lead to reordering and performance will drop.
```
nicorusti, line 1829:
```suggestion
Streams can be used within a session to distinguish between different packet flows. A single stream has a single reassembly queue and SHOULD be processed by a single core. Packets in a stream SHOULD NOT be distributed over multiple SCION paths to avoid reordering and therefore lower performance.
```

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,86 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+
+5. The remote SIG receives the SCION packet and decapsulates the original IP packet. It then forwards the packet to the final IP destination using standard local IP routing.
+
+## SIG Framing
+IP packets are encapsulated over SCION/UDP into SIG frames. While in principle, a pair of SIGs may use other tunneling protocols, existing deployments leverage SIG framing as described here.
+TODO: @oncilla @shitz could you say something about why this framing and not existing protocol? Is it because of MTU?
+
+There may be multiple IP packets in a single SIG frame, and a single IP packet may be split into multiple SIG frames.
+A pair of SIGs establishes a SIG tunneling session.
+TODO: how are the sessions established? Is a session mapped to a single path?
```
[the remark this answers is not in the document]
nicorusti: THanks a lot, @knmeynell added a derivative of this text to the draft

draft-dekater-scion-dataplane.md
```diff
@@ -1806,14 +1805,84 @@ However, the path choice of the endpoint may possibly be exploited by an attacke
...
+- `Session ID` (8 bits) identifies a tunneling session between a pair of SIGs.
+- `Index` (16 bits) is the byte offset of the first beginning of an IP packet within the payload. If no IP packet starts in the payload, e.g. if the frame contains only the middle or trailing part of an IP packet, the field MUST be set to 0xFFFF.
+- `Reserved` (12 bits): it MUST be set to zero.
+- `Stream ID` (20 bits), along with the session, it identifies a unique sequence of SIG frames. Frames from the same stream are, on the remote SIG, put into the same reassembly queue. There may be multiple streams per session.
+- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames can be used to reassemble IP packets split among multiple frames. In the rare event that a frame is delayed or lost and a frame other than the next expected one is received, any previously received frames pertaining to a yet incomplete IP packet, are discarded. Subsequent frames are discarded until one carries the beginning of an IP packet. Late frames (with a sequence number older than any non-discarded ones) are discarded as they necessarily belong to an already discarded IP packet."
```
nicorusti, line 1875:
```suggestion
- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames are used to reassemble IP packets split among multiple frames.

SIGs MAY reorder received frames within a stream. Choice of frame reassembly window is left to implementations.
```

jiceatscion:
For simplicity of the process, I accepted suggestions. I think this is good enough to merge (even if we leave it at that). We can do another pass after that.

nicorusti (approved):
I left a couple of comments, I think after this this is good to merge  (sorry, one more suggestion for you @jiceatscion 🥲 )

jiceatscion:
Accepted. In the next pass I will mention that the egress SIG is free to buffer packets for re-ordering, or discard; as a unilateral decision.

PR #70, opened 2024-10-09
Checksum rewording

### LINKED ISSUE #47: Upper layer checksums
The upper layer checksum raises questions.. However this is used in the current implementation. 

- [x] Can we adapt the text in the draft?
- [ ] Respond to Harald Alvestrand and Mike Mc Bride

> Harald Alvestrand
SCION keeps accommodating the “checksum across layers 3 and 4”. This was a bad idea in 1975, and keeps being a bad idea now. The idea should (in my opinion) be relegated to gateways and kept out of the core.

> The payload of a packet can be anything (does not affect SCION operation), but the header contains an integer saying which content it is, with the defined values being “UDP”, “TCP”, and some numbers for internal use. The process for constructing an IP packet from the SCION packet (or vice versa) is not described, but presumably involves mangling checksums - -dataplane describes a “virtual header” for constructing the checksum of things that require it, which implies that a gateway to non-SCION IP would mangle checksums inside the UDP or TCP packet.
(There are references in -overview to docs des 

>  Mike McBride' <mmcbride7@gmail.com>
I discussed potential scion checksum issues (like why have them) with Eliot and understand that they’ve been discussed with scion authors so won’t get into that.

nicorusti:
Resolves #47

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -866,7 +866,8 @@ The PadN option is used to insert two or more bytes of padding into the `Options
 
 ## Pseudo Header for Upper-Layer Checksum {#pseudo}
 
-Any transport or other upper-layer protocol that includes addresses from the SCION header in the checksum computation MUST use the following pseudo header:
+The SCION data plane does not provide payload integrity protection, as further clarified in [](#payload-integrity).
+Should any transport or other upper-layer protocols compute a checksum of the SCION header, then they SHOULD use the following pseudo header:
 
 ~~~~
  0                   1                   2                   3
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-dataplane.md
```diff
@@ -896,6 +897,7 @@ Any transport or other upper-layer protocol that includes addresses from the SCI
 - `Upper-Layer Packet Length`: The length of the upper-layer header and data. Some upper-layer protocols define headers that carry the length information explicitly (e.g., UDP). This information is used as the upper-layer packet length in the pseudo header for these protocols. The remaining protocols, which do not carry the length information directly, use the value from the `PayloadLen` field in the SCION common header, minus the sum of the extension header lengths.
 - `Next Header`: The protocol identifier associated with the upper-layer protocol (e.g., 17 for UDP - see also [](#protnum)). This field can differ from the `NextHdr` field in the SCION common header, if extensions are present.
 
+This pseudo-header is used in current implementations of UDP/SCION on top of SCION. However, checksums across layers are not recommended and should be re-evaluated in future evolutions.
```
jiceatscion, line 900: I would suggest a slightly different turn of phrase, just for grammar:
This pseudo-header is used in current implementations of UDP on top of SCION. However, as checksums across layers are not recommended, this should be re-evaluated in future revisions.

jiceatscion (approved):
Suggestion non-binding; as usual.

PR #74, opened 2024-10-15
remove mention of multiple ISDs

### LINKED ISSUE #37: Clarify how routers use BFD
We currently mention BFD in [2.2.3.1. ](https://www.ietf.org/archive/id/draft-dekater-scion-dataplane-02.html#section-2.2.3.1)[Empty Path Type](https://www.ietf.org/archive/id/draft-dekater-scion-dataplane-02.html#name-empty-path-type): 
> One use case of the Empty path type lies in the context of link-failure detection. To this end, SCION uses the Bidirectional Forwarding Detection (BFD) protocol ([[RFC5880](https://www.ietf.org/archive/id/draft-dekater-scion-dataplane-02.html#RFC5880)] and [[RFC5881](https://www.ietf.org/archive/id/draft-dekater-scion-dataplane-02.html#RFC5881)]). BFD is a protocol intended to detect faults in the bidirectional path between two forwarding engines, with typically very low latency. It operates independently of media, data protocols, and routing protocols. SCION uses the Empty path type, together with OneHopPath path type, to bootstrap BFD within SCION. (For more information on the OneHopPath path type, see [Section 2.2.3.3](https://www.ietf.org/archive/id/draft-dekater-scion-dataplane-02.html#onehop).)

However, we don't say much about how it works.
- How much should be in the draft? (is it necessary to add details, or could we leave it open)?
- Why do routers have to use something like BFD? What do they do with signals from BFD?
- How is BFD on top of SCION working? (currently described in https://docs.scion.org/en/latest/protocols/bfd.html )

### LINKED ISSUE #73: Remove mention of AS in multiple ISDs
From Harald https://github.com/scionassociation/standards/issues/105 
> A couple of things are described either very briefly or inconsistently - for instance, the claim is made that “one AS can be in multiple ISDs” - yet the addressing consists of a 16 bit ISD number and a 48 bit AS number - it would seem that one AS in multiple ISDs would have to operate as if it was two ASes.

```
      the up and down segments do not need to originate from the same
      core AS and can even be in different ISDs (if the AS at the
      intersection is part of multiple ISDs).
```

Background on this: https://github.com/scionproto/scion/issues/4293

nicorusti:
Resolves #73

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -384,7 +384,7 @@ Valid path segment combinations:
   - **Immediate combination** (Cases 2a, 2b in {{figure-1}}): The last AS on the up segment (which is necessarily a core AS) is the same as the first AS on the down segment. In this case, a simple combination of up and down segments creates a valid forwarding path. In Case 2b, only one segment is required.
 
 - **Peering shortcut** (Cases 3a and 3b): A peering link exists between the up and down segment, and extraneous path segments to the core are cut off. Note that the up and down segments do not need to originate from the same core AS and the peering link could also be traversing to a different ISD.
-- **AS shortcut** (Cases 4a and 4b): The up and down segments intersect at a non-core AS below the ISD core, thus creating a shortcut. In this case, a shorter path is made possible by removing the extraneous part of the path to the core. Note that the up and down segments do not need to originate from the same core AS and can even be in different ISDs.
+- **AS shortcut** (Cases 4a and 4b): The up and down segments intersect at a non-core AS below the ISD core, thus creating a shortcut. In this case, a shorter path is made possible by removing the extraneous part of the path to the core. Note that the up and down segments do not need to originate from the same core AS.
 - **On-path** (Case 5): In the case where the source's up segment contains the destination AS or the destination's down segment contains the source AS, a single segment is sufficient to construct a forwarding path. Again, no core AS is on the final path.
 
 ## Path Authorization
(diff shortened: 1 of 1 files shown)


jiceatscion (approved):
...some readers who pay attention might ask how it happens that two segments in different ISDs intersect at one AS... If you accept leaving the response as an exercise to the reader, then all right.

nicorusti:
Coming from here: 
Screenshot 2024-10-15 at 11 42 25

draft-dekater-scion-dataplane.md
```diff
@@ -384,7 +384,7 @@ Valid path segment combinations:
...
 
 - **Peering shortcut** (Cases 3a and 3b): A peering link exists between the up and down segment, and extraneous path segments to the core are cut off. Note that the up and down segments do not need to originate from the same core AS and the peering link could also be traversing to a different ISD.
-- **AS shortcut** (Cases 4a and 4b): The up and down segments intersect at a non-core AS below the ISD core, thus creating a shortcut. In this case, a shorter path is made possible by removing the extraneous part of the path to the core. Note that the up and down segments do not need to originate from the same core AS and can even be in different ISDs (if the AS at the intersection is part of multiple ISDs).
+- **AS shortcut** (Cases 4a and 4b): The up and down segments intersect at a non-core AS below the ISD core, thus creating a shortcut. In this case, a shorter path is made possible by removing the extraneous part of the path to the core. Note that the up and down segments do not need to originate from the same core AS and can even be in different ISDs.
```
nicorusti, line 387:
```suggestion
- **AS shortcut** (Cases 4a and 4b): The up and down segments intersect at a non-core AS below the ISD core, thus creating a shortcut. In this case, a shorter path is made possible by removing the extraneous part of the path to the core.
```

PR #75, opened 2024-10-15
sig: clarify what's required for compatibility and what is unspecified

### LINKED ISSUE #40: Specify SCION-IP Gateways - outline

jiceatscion:
Fixes #40

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -1855,22 +1855,32 @@ Tunneling IP traffic over SCION requires a pair of SIGs and it involves the foll
 
 2. The IP packet reaches a SIG in the sender’s network via standard IP routing.
 
-3. Based on the destination IP address, the source SIG determines the destination SIG's ISD-AS endpoint address. To achieve this, SIGs may be pre-configured with a static IP prefix to remote SIG SCION address mappings. Alternatively, a pair of SIGs may be configured to speak a dynamic routing protocol between each other. The choice of protocol is left up to implementors and is outside of the scope of this document.
+3. Based on the destination IP address, the source (ingress) SIG determines the destination (egress) SIG's ISD-AS endpoint address. To achieve this, SIGs may be pre-configured with a static IP prefix to remote SIG SCION address mappings. Alternatively, multiple SIGs may be configured to speak a dynamic routing protocol between each other. The choice of protocol is left up to implementors and is outside of the scope of this document. Whether an egress SIG accepts or drops traffic from a given ingress SIG is up to the policies of the receiver's operator. The control mechanisms are implementation defined.
 
-4. The SIG encapsulates the original IP packet within a SCION packet and sends it to the remote SIG. If necessary, the source SIG performs SCION path lookups and selects a SCION path to the destination SIG.
+4. The ingress SIG encapsulates the original IP packet within a SCION packet and sends it to the egress SIG. If necessary, the ingress SIG performs SCION path lookups and selects a SCION path to the egress SIG.
 
-5. The remote SIG receives the SCION packet and decapsulates the original IP packet. It then forwards the packet to the final IP destination using standard IP routing.
+5. The egress SIG receives the SCION packet and decapsulates the original IP packet. It then forwards the packet to the final IP destination using standard IP routing.
 
 ## SIG Framing
 
-IP packets are encapsulated over SCION/UDP into SIG frames. Whilst in principle, a pair of SIGs may use other tunneling protocols, existing deployments use SIG framing as described here. This is to provide independence from the underlying SCION path MTU which can increase and decrease over time, to provide fast detection of packet loss and subsequent recovery of decapsulation for packets that weren't lost, and support for multiple streams within a framing session such that the streams can be distributed to separate cores.
+IP packets are encapsulated over SCION/UDP into SIG frames. Whilst in principle, a given pair of SIGs may use any tunneling protocol, existing deployments use the SIG framing as described here. This protocol is designed to:
+
+- provide independence from the underlying SCION path MTU which can increase and decrease over time.
+- provide fast detection of packet loss and subsequent recovery of decapsulation for packets that weren't lost.
+- support for multiple streams within a framing session such that independent packet sequences be tunneled in parallel.
 
 There may be multiple IP packets in a single SIG frame, and a single IP packet may be split into multiple SIG frames.
-A source SIG unidirectionally establishes a SIG tunneling session with a destination SIG. A session may be mapped to one or multiple SCION paths, depending on implementors.
 
-Streams can be used within a session to distinguish between different packet flows. A single stream has a single reassembly queue and SHOULD be processed by a single core. Packets in a stream SHOULD NOT be distributed over multiple SCION paths to avoid reordering and therefore lower performance.
+The ingress SIG initiates unidirectional packet flows with the egress SIG simply by sending the corresponding SIG frames. There is no peamble. The egress SIG, should it accept the traffic, instanciates the necessary resources on-demand to process each flow. Each such flow forms an independent sequence of packets (a stream) ordered by an incrementing sequence number. Between a given SIG ingress/egress pair, a (session ID, stream ID) pair uniquely identifies a stream.
 
-Each SIG frame has a sequence number that is used by the egress SIG to reassemble the encapsulated IP packets within a stream.
+To preserve performance, IP packets encapsulated in a single stream SHOULD NOT be re-ordered. To that end:
+
+- The ingress SIG SHOULD encapsulate IP packets that cannot be proven independent (for example with the same source, destination and traffic class) in the same stream.
+- The ingress SIG SHOULD encapsulate IP packets to a given stream in the order in which they were received.
+- The ingress SIG SHOULD send all frames of a given stream over a the same SCION path.
+- The egress SIG SHOULD reassemble and forward packets from each stream, ordered by frame sequence number and packet within each frame.
+
+The session ID part of the (session ID, stream ID) pair is used to indicate traffic priority grouping. The egress SIG MAY dedicate processing resources to each session rather than to each individual stream.
 
 ~~~~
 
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-dataplane.md
```diff
@@ -1855,22 +1855,32 @@ Tunneling IP traffic over SCION requires a pair of SIGs and it involves the foll
...
+- support for multiple streams within a framing session such that independent packet sequences be tunneled in parallel.
 
 There may be multiple IP packets in a single SIG frame, and a single IP packet may be split into multiple SIG frames.
-A source SIG unidirectionally establishes a SIG tunneling session with a destination SIG. A session may be mapped to one or multiple SCION paths, depending on implementors.
 
-Streams can be used within a session to distinguish between different packet flows. A single stream has a single reassembly queue and SHOULD be processed by a single core. Packets in a stream SHOULD NOT be distributed over multiple SCION paths to avoid reordering and therefore lower performance.
+The ingress SIG initiates unidirectional packet flows with the egress SIG simply by sending the corresponding SIG frames. There is no peamble. The egress SIG, should it accept the traffic, instanciates the necessary resources on-demand to process each flow. Each such flow forms an independent sequence of packets (a stream) ordered by an incrementing sequence number. Between a given SIG ingress/egress pair, a (session ID, stream ID) pair uniquely identifies a stream.
```
nicorusti, line 1874:
```suggestion
The ingress SIG initiates unidirectional packet streams with the egress SIG simply by sending the corresponding SIG frames. There is no handshake. The egress SIG, should it accept the traffic, instanciates the necessary resources on-demand to process each stream. Each such  stream consists of an independent sequence of packets ordered by an incrementing sequence number. Between a given SIG ingress/egress pair, a (session ID, stream ID) pair uniquely identifies a stream.
```
I think your version uses a lot the word "flow" to talk about a stream. How about we just talk about streams? 
Also., I'd prefer handshake rather than p**R**eamble

draft-dekater-scion-dataplane.md
```diff
@@ -1855,22 +1855,32 @@ Tunneling IP traffic over SCION requires a pair of SIGs and it involves the foll
...
+
+To preserve performance, IP packets encapsulated in a single stream SHOULD NOT be re-ordered. To that end:
```
nicorusti, line 1876: >  IP packets encapsulated in a single stream SHOULD NOT be re-ordered. 

This sentence is a bit ambiguous, it feels like the destination SIG should not do any reordering

 ```suggestion
To preserve performance,  SIGs SHOULD take the following measures to minimize chances of re-ordering:
```

File: draft-dekater-scion-dataplane.md [hunk 3]
```diff
@@ -1855,22 +1855,32 @@ Tunneling IP traffic over SCION requires a pair of SIGs and it involves the foll
...
+
+- The ingress SIG SHOULD encapsulate IP packets that cannot be proven independent (for example with the same source, destination and traffic class) in the same stream.
+- The ingress SIG SHOULD encapsulate IP packets to a given stream in the order in which they were received.
+- The ingress SIG SHOULD send all frames of a given stream over a the same SCION path.
+- The egress SIG SHOULD reassemble and forward packets from each stream, ordered by frame sequence number and packet within each frame.
```
nicorusti, line 1881:
```suggestion
- The ingress SIG SHOULD encapsulate IP packets that cannot be proven independent (e.g., with the same  IP 6-tuple) in the same stream.
- The ingress SIG SHOULD encapsulate IP packets to a given stream in the order in which they were received.
- The ingress SIG SHOULD send all frames of a given stream over the same SCION path.
- The egress SIG SHOULD reassemble and forward packets from each stream, ordered by frame sequence number and packet within each frame.
```
We don't describe what traffic class we are referring to (the SCION header traffic class? The IP header one?). Id therefore remove the mention of traffic class

nicorusti (changes requested):
Thanks a lot, I left a few suggestions. 
I don't understand what "priority grouping" means, maybe we can quickly discuss

draft-dekater-scion-dataplane.md
```diff
@@ -1913,17 +1923,17 @@ All fields within SIG Frame Header are in network byte order.
...
-SIGs MAY reorder received frames within a stream. Choice of frame reassembly window is left to implementations.
+A SIG MAY drop frames. Sorting frames received out-of-order by the egress SIG is optional. The egress SIG SHOULD drop frames from a stream if unable to perform the sequence re-assembly.
 
 The Session ID and Stream ID are chosen by the sender but the tuple MUST be unique within a session.
 
 
 ## SIG Frame Payload
 
-The SIG frame payload may contain multiple IPv4 or IPv6 packets, or parts thereof. No other types of packets can be encapsulated and the packets are placed directly after one another with no padding.
+The SIG frame payload may contain multiple IPv4 or IPv6 packets, or parts thereof. No other types of packets can be encapsulated and the packets are placed directly after one another with no padding. Handling of multicast is not covered by this specification and it is left as future work.
```
nicorusti, line 1936: @jiceatscion JFYI Eliot recommended to mention multicast tunneling is not in scope. I added this

jiceatscion: lgtm

draft-dekater-scion-dataplane.md
```diff
@@ -1913,10 +1923,10 @@ All fields within SIG Frame Header are in network byte order.
...
-- `Stream ID` (20 bits), along with the session, it identifies a unique sequence of SIG frames. Frames from the same stream are, on the remote SIG, put into the same reassembly queue. There may be multiple streams per session.
-- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames are used to reassemble IP packets split among multiple frames.
+- `Stream ID` (20 bits), along with the session, it identifies a unique sequence of SIG frames. Frames from the same stream are, on the egress SIG, put into the same reassembly queue. There may be multiple streams per session.
+- `Sequence Number` (64 bits) indicates the position of the frame within a stream. Consecutive frames of a given stream have consecutive sequence numbers. IP packets split among multiple frames are re-assembled by concatenating the payloads of consecutive frames.
 
-SIGs MAY reorder received frames within a stream. Choice of frame reassembly window is left to implementations.
+A SIG MAY drop frames. Sorting frames received out-of-order by the egress SIG is optional. The egress SIG SHOULD drop frames from a stream if unable to perform the sequence re-assembly.
```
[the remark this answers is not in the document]
jiceatscion: My reason for this phrasing is to avoid ambiguity. Frames do have an order, defined by the sequence number. I want to be clear that the "re-ordering" we're talking about is to process them into *that* order and not changing it. Apparently that doesn't work. But I am not satisfied with just saying "reordering" I think it is not precise and relies on the audience guessing the meaning. May be the more specific "buffering frames received out-of-order is optional" would work?

draft-dekater-scion-dataplane.md
```diff
@@ -1855,22 +1855,32 @@ Tunneling IP traffic over SCION requires a pair of SIGs and it involves the foll
...
+
+To preserve performance, IP packets encapsulated in a single stream SHOULD NOT be re-ordered. To that end:
+
+- The ingress SIG SHOULD encapsulate IP packets that cannot be proven independent (for example with the same source, destination and traffic class) in the same stream.
+- The ingress SIG SHOULD encapsulate IP packets to a given stream in the order in which they were received.
+- The ingress SIG SHOULD send all frames of a given stream over a the same SCION path.
+- The egress SIG SHOULD reassemble and forward packets from each stream, ordered by frame sequence number and packet within each frame.
 
-Each SIG frame has a sequence number that is used by the egress SIG to reassemble the encapsulated IP packets within a stream.
+The session ID part of the (session ID, stream ID) pair is used to indicate traffic priority grouping. The egress SIG MAY dedicate processing resources to each session rather than to each individual stream.
```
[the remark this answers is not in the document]
jiceatscion: Actually the grouping does not guarantee any effective priority. The egress SIG has no idea what the various sessions are for. All it does is to give each session its own task, thereby preventing head-of-line blocking. The priorities arise from the relative loads placed by the sender on the sessions. In the OS impl, streams within a session have separate queues but they do not have separate tasks. That is the *only* effect of sessions that I could find in the impl, so I thought I'd use it as a way to explain why the protocol has an explicit ID pair to identify a stream: it enables this kind of hint from the ingress to the egress stream. Depending on implementation it may have an impact (e.g. in the OS impl) or not (e.g. if the egress SIG gives a core to every stream or has a single core running them all). This is basically the protocol enabling some subtle QoS behaviours without being too complicated. I am not sure how to explain that. If I don't people will ask what the two-level stream ID is for. I agree that "priority grouping" is crappy. Suggestions welcome for an alternate term.

draft-dekater-scion-dataplane.md
jiceatscion: I was quite deliberate in avoiding the term "stream" until I defined it. That is because the protocol identifies what we have to call a stream by "session ID" and "stream ID". So we have the so common and irritating case of a stream ID not being the ID of a stream. My phrasing was attempt at containing the issue.

jiceatscion: I do agree with your suggestion of replacing "preamble" with "handshake".

nicorusti: Ok replaced with handshake

draft-dekater-scion-dataplane.md
nicorusti: Ok I like your wording, added it

> To preserve performance, IP packets encapsulated in a single stream SHOULD leave the egress SIG in the order in which they entered it. To that end:

draft-dekater-scion-dataplane.md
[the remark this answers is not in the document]
nicorusti: Sounds good, went for 
>  Buffering frames received out-of-order by the egress SIG is optional.

nicorusti (approved):

PR #76, opened 2024-10-15
Suggested change to remove specific reference to CPU cores

knmeynell:
This is less implementation specific.


nicorusti:
Conflicts with #75, let's review that one first and then see how we remove mentions of CPUs (I think JC already did it)

PR #79, opened 2024-10-16
Removed SIG protocol description

knmeynell:
Now contains brief SIG description and note that IP tunnelling is out-of-scope of the draft.


nicorusti (commented):
The text sounds good, however, do you think it could fit as a subsection? I think it would fit nicely below MTU & fragmentation: 
```
 3.  Life of a SCION Data Packet
     3.1.  Description
     3.2.  Creating an End-to-End SCION Forwarding Path
     3.3.  Step-by-Step Explanation
     3.4.  MTU
     3.5.  Packet Fragmentation
 ```
 
 However, both the fragmentation and MTU sections are not really part of life of a packet. Maybe we can create a new section # Deployment considerations (including MTU, fragmentation, SIG)? 

P.S. I learned a new word, outwith

nicorusti:
JFYI I updated the definition of endpoint also in the CP draft to not mention SIG in https://github.com/scionassociation/scion-cp_I-D/pull/60

nicorusti (approved):

PR #87, opened 2024-11-12
scmp: export SCMP to controlplane draft

### LINKED ISSUE #86: hopfield expiration: the calculation is wrong
The formula for the absolute expiration time of a hop field is wrong. It implies that the duration unit is a 256th of an hour. In reality it is a 256th of a day. We also do not actually describe the unit; we just throw it in the formula.

jiceatscion:
Also, in passing:
* fix the description of the hopfield expiration time field
* delete left-over text fragment "(see [])".

Resolves: #84 
Resolves: #86

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -763,7 +763,7 @@ The 12-byte Hop Field (``HopField``) has the following format:
 
 The Ingress Router (respectively Egress Router) is the router owning the Ingress interface (respectively, Egress interface) when the packet is traveling in the *construction direction* of the path segment (i.e. the direction of beaconing). When the packet is traveling in the opposite direction, the meanings are reversed.
 
-Router alert flags work similarly to {{RFC2711}} and allow a sender to address a specific router on the path without knowing its address. Processing the L4 payload in the packet means that the router will treat the payload of the packet as a message to itself and parse it according to the value of the `NextHdr` field. Such messages include Traceroute Requests (see {{I-D.dekater-scion-controlplane}} section "SCMP/Traceroute Request")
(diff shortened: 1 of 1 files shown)


nicorusti (approved):
LGTM (I fixed a couple of minor things)

nicorusti (commented):
Should this section go to the CP?

draft-dekater-scion-dataplane.md
```diff
@@ -1804,7 +1424,7 @@ A SCION router SHOULD accept BFD connections from its peers and SHOULD attempt t
 
 ## Link Failure Notification - SCMP {#link-down-notification}
 
-In SCION, an intermediate router cannot change the path followed by a packet, only the source endpoint can chose a different path. Therefore, to enable fast recovery, a router SHOULD signal forwarding failures to the source, via a [SCMP notification](#scmp-notification). This allows the source endpoint to quickly switch to a different path.
+In SCION, an intermediate router cannot change the path followed by a packet, only the source endpoint can chose a different path. Therefore, to enable fast recovery, a router SHOULD signal forwarding failures to the source, via a SCMP notification (see {{I-D.dekater-scion-controlplane}} section "SCMP/Error Messages"). This allows the source endpoint to quickly switch to a different path.
 
 Sending an SCMP error notification is OPTIONAL. Endpoints should therefore implement additional mechanisms to validate or detect link down signals. To reduce exposure to denial-of-service attacks, SCION routers SHOULD employ rate limiting when sending recommended SCMP notifications (especially identical ones). Rate limit policies are up to each AS' administrator.
```
nicorusti, line 1429: Should this section go to the CP?

jiceatscion: I don't think so. Just because the control plane provides the SCMP service (by some definition of provides, anyway), doesn't mean that everything that makes use of it becomes part of the control service, does it?

As you seem to let me make that choice, I'm merging.

PR #96, opened 2025-06-13
Terminology Updates

knmeynell:


nicorusti:
Wording example from https://datatracker.ietf.org/doc/html/rfc4861
```
unspecified address
               - a reserved address value that indicates the lack of an
                 address (e.g., the address is unknown).  It is never
                 used as a destination address, but may be used as a
                 source address if the sender does not (yet) know its
                 own address (e.g., while verifying an address is unused
                 during stateless address autoconfiguration [[ADDRCONF](https://datatracker.ietf.org/doc/html/rfc4861#ref-ADDRCONF)]).
                 The unspecified address has a value of 0:0:0:0:0:0:0:0.
                 ```

nicorusti (approved):

PR #95, opened 2025-06-03
Updated Figures and added monodraw-generated SVGs

knmeynell:
Resolves #94 

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-dp_I-D/draft-dekater-scion-dataplane.txt&url_2=https://scionassociation.github.io/scion-dp_I-D/New_Diagrams/draft-dekater-scion-dataplane.txt)

PR #97, opened 2025-06-30
One-Hop Path Type: remove figure 12

knmeynell:
Follow up to PR #95 to add one-hop path type text


nicorusti:
ZDo you remember why we are changing this? Just to remove the figure, right? 

How about this wording? 
```
The `OneHopPath` path type (`PathType=2`) is currently used to bootstrap beaconing between neighboring ASes. This is necessary as neighbor ASes do not have a forwarding path before beaconing is started.

A one-hop path has exactly one Info Field and two Hop Fields. The second Hop Field is created by the ingress SCION border router of the neighboring AS while processing the one-hop path. The appropriate Hop Field can be processed by a border router based on the source and destination address. In this context, the following rules apply:

- At the source endpoint AS, *CurrHF := 0*.
- At the destination endpoint AS, *CurrHF := 1*.


Any entity with access to the forwarding key of the source endpoint AS can create a valid info and Hop Field as described in [](#inffield) and [](#hopfld), respectively.

Upon receiving a packet containing a one-hop path, the ingress border router of the destination AS fills in the `ConsIngress` field in the second Hop Field of the one-hop path with the ingress interface ID. It sets the `ConsEgress` field to an invalid value (e.g. unspecified value 0), ensuring the path cannot be used beyond the destination AS. Then it calculates and appends the appropriate MAC for the Hop Field.
```

PR #117, opened 2025-10-31
Clarifications - Brian Trammel's review

### LINKED ISSUE #114: Brian Trammell's review DP
The SCION Data Plane draft specifies the mechanism for end-to-end packet forwarding within the SCION architecture. The document's core purpose is to define how data packets combine authorized path segments (obtained via the Control Plane) into complete paths for transmission. This component is interdependent with the SCION Control Plane, which handles path discovery and secure dissemination, and the Control Plane PKI, which provides the necessary cryptographic basis for trust and scalability based on Isolation Domains (ISDs).

Why a new DP at layer 3? Don't we have at least two, plus a host of shim and transition layers, and ways to use them to do all kinds of traffic engineering, at least in intradomain contexts? The argument for another one here lies in said CP interdependence: the hop fields (HF) generated by the CP could be wedged into other encapsulations, but during the development and initial deployment of the protocol, a bespoke DP makes sense, and for the purposes of this document (describing SCION as is for posterity, as well as as a basis for future RG and
WG work derived from the protocol's current state) is perfectly fine.

However, this is my primary comment on the document as a whole: moving forward, the necessity of defining an entirely novel network layer data plane bears closer scrutiny. While Segment Routing over IPv6 (SRv6), for example, is generally constrained to intradomain applications by its architectural assumptions, its segments look a lot like SCION hop fields, and I suspect exploring this parallelism further could be fruitful. Since SCION is primarily focused on interdomain routing, treating each AS as an opaque problem presumably already solved by MPLS, it also makes sense to look at other existing methods of encapsulation for gluing MPLS fabrics together at peering and transit links, all of which would presumably less disruptive to existing border silicon and operational practices to deploy than a shiny new DP spec.

Indeed, SCION makes the exact layout of the DP packets less important to total forwarding performance: since each HF specifies the ingress and egress Interface IDs required for traversal through an individual Autonomous System (AS). This contrasts sharply with IP, where forwarding is based on destination address lookups, allowing SCION border routers to operate without consulting large inter-domain forwarding tables.

_KM - I don't think anything needs to be done with the above points._ 

## Technical Issues

- Section 1.3.3 Configuration: the requirement to provide out of band configuration of interface ID to intradomain routing directive (mpls label stack, intranet IP address, optical switch port, etc) is perhaps unavoidable given the specified size of the HF, but I see this as a potentially major impediment to SCION deployment in large domains. The operations model assumes a convergent eventual consistency to the configurations of all devices in a domain, which is perhaps optimistic. Future standardization of a SCION protocol stack will need to carefully consider the operational considerations of HF to intradomain addressing mappings.

- Section 2.3.3.4: This is basically the same issue: defining the hop fields such that one could build configuration-free encrypted versions of the internal MPLS label (stack), for instance, would significantly improve configurability and operability.

_KM - the above would seem to be philosophical questions._

## Editorial Issues

- [x] The use of binary literals in Table 3 is confusing; I'd prefer to see these as separate fields rather than a nybble literal.

- [x] Section 2.3.2.1, "C (urrINF)": this was a little confusing when I first saw it; the name of the field in the packet format diagram and the name in the description would be easier to read if they match. I'd suggest using more explicit language in the description to note that C means "current info".

- [x] Section 2.5: "However, as checksums across layers are not recommended, this should be re-evaluated in future revisions." I'd consider making this stronger ("we don't recommend doing this going forward").

- [x] Section 5.3: I don't think "outwith" is a word; "outside"?
_KM - Outwith is certainly a [word](https://www.oed.com/dictionary/outwith_prep?tab=factsheet#32840462) with slightly different connotations compared to 'outside'. However, I've changed this to "outside of scope"._ 

- [x] Section 1.1. Terminology: ISD: while a lot of the initial interest in SCION has come the fact that ISD boundaries look a lot like catenet boundaries, and there is a philosophy of the internet that is interested in the network's boundaries aligning with jurisdictional boundaries, and one could even posit that this particular ship has sailed with the growth of the "splinternet", note that "a possible model is for ISDs to be formed
along national boundaries" is probably not an aside I would add to the terminology section of a document attempting to sell the IETF community on an idea. This is only one possible model, and it would be a shame for people to dismiss ISDs as a tool because the implications of one of its applications is problematic.
_KM - Yes, I don't feel we should be explicit about use cases in an Internet Draft and I was never comfortable with the inclusion of this._

- [x] Section 2. Header spec: is 32-bit alignment still relevant for the devices that would be running SCION (I'd assume 8- or 64- would be?) or is this simply a nod to history? --> Yes it is relevant. We can say "This alignment is used to compute header length based on the `HdrLen` field (reference to 2.1.  Common Header)

nicorusti:
Resolves #114 

<br class="Apple-interchange-newline">[SCION DP](https://scionassociation.github.io/scion-dp_I-D/brian-review-v2/draft-dekater-scion-dataplane.html)	[plain text](https://scionassociation.github.io/scion-dp_I-D/brian-review-v2/draft-dekater-scion-dataplane.txt)	[diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-dp_I-D/draft-dekater-scion-dataplane.txt&url_2=https://scionassociation.github.io/scion-dp_I-D/brian-review-v2/draft-dekater-scion-dataplane.txt)

### DIFF
--- a/draft-dekater-scion-dataplane.md
+++ b/draft-dekater-scion-dataplane.md
@@ -367,7 +367,7 @@ Border routers require mappings from SCION Interface IDs to underlay addresses a
 - For the routers that do not manage the interface: the address of the intra-domain protocol on the router that does.
 - The algorithm used to compute the [Hop Field MAC](#hf-mac-overview) which must be the same as that used by the Control Services within the AS.
 
-In order to forward traffic to a service endpoint address (`DT/DS` == 0b01 in the [common header](#common-header)), a border router translates the service number into a specific destination address. The method used to accomplish the translation is not defined by this document and is only dependent on the implementation and the choices of each AS's administrator. In current practice this is accomplished by way of a configuration file.
+In order to forward traffic to a service endpoint address (`DT/DS` as per {{table-3}}), a border router translates the service number into a specific destination address. The method used to accomplish the translation is not defined by this document and is only dependent on the implementation and the choices of each AS's administrator. In current practice this is accomplished by way of a configuration file.
 
 **Note:** The current SCION implementation runs over the UDP/IP protocol. However, the use of other lower layers protocols is possible.
 
@@ -555,13 +555,13 @@ The SCION common header has the following packet format:
 | 3           | 16 bytes       |
 {: #table-2 title="Address length values"}
 
-| Length (bytes) | Type | DT/DL or ST/SL encoding | Conventional Use |
-|----------------+------+-------------------------+------------------|
-| 4              | 0    | 0b0000                  | IPv4             |
-| 4              | 1    | 0b0100                  | Service          |
-| 16             | 0    | 0b0011                  | IPv6             |
-| other          |      |                         | Unassigned       |
-{: #table-3 title="Allocations of type values to length values"}
+| Type (DT/ST) | Length (DL/SL) | Conventional Use |
+|--------------+----------------+------------------|
+| 0            | 0              | IPv4             |
+| 0            | 3              | IPv6             |
+| 1            | 0              | Service          |
+| other        | other          | Unassigned       |
+{: #table-3 title="Allocations of length and type combinations"}
 
 A service address designates a set of endpoint addresses rather than a singular one. A packet addressed to a service is redirected to any one endpoint address that is known to be part of the set. {{table-4}} lists the known services.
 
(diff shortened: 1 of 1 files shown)


nicorusti:
@knmeynell feel free to merge if wording is ok

